PR #11, opened 2017-01-27
New HTML documentation and a port of existing HTML rules

deadlight:
### Done
Copied existing HTML rules from the "Syntax" section of the old [guidelines style guide](https://docs.google.com/document/d/1jvIKgLH-UJNPt1yGw804qUXoj6N6Fsd5_TA2is7K2XI/edit#heading=h.n2khz4x2pums) and added the recent rule of two-space indentation

PR #13, opened 2017-01-27
CONTRIBUTING.md: Rules for approving pull requests

nottrobin:
Define the threshold for and practice around approving a pull request.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,3 +7,15 @@ Anyone is welcome to suggest improvements in one of two ways:
 
 - Start a discussion by submitting [a new issue](https://github.com/ubuntudesign/webteam-practices/issues/new)
 - Suggest a change by submitting a pull request ([how?](https://help.github.com/articles/about-pull-requests/))
+
+Approving a pull request
+---
+
+Before a pull request is approved and merged:
+
+- A reasonable effort should be made to bring it to the attention of the whole team
+- It should also have been :+1:d by at least 2 team members
+- It should have been open for at least 24 hours
+
+Once it is merged, it should be brought up in the next relevant team meeting to ensure the whole team is aware of the change.
+


nottrobin:
@anthonydillon said:

> I don't think we need to move too quickly on these reviews. I would like to say at least 2 +1's and have the PR up for more than 24 hours so everyone gets a chance to review.

nottrobin:
@deadlight rebutted:

> @anthonydillon I think that it would be better to get them closed relatively quickly and have any changes issued as PRs. Especially useful for times like right now where I have more changes waiting on a PR approval.

nottrobin:
I like @anthonydillon's idea. I think we should be slow about making policy. Once something is merged, it's official. It can be pointed to in arguments to shut them down. It is The Way We Work from then on.

WillMoggridge:
Going forward, I definitely prefer that it is left for 24 hours. As it stands right now, while we flesh out the repository a bit, I don't mind so much.

nottrobin:
@anthonydillon said:

> But we will miss people opinions and once it lands becomes approved and may be difficult to change with another PR.

nottrobin:
I've updated the wording to include @anthonydillon's suggestion (not yet agreed)

anthonydillon:
Agree @WillMoggridge

anthonydillon:
But we need to transition from fast to slow and I'm not sure how and who makes that call. It also devalued the current content because we are in our "fast phase"

deadlight:
Agreed but the current content is binding even if we are moving fast. It can't be devalued, just slowly changed.

nottrobin:
Yes I think that is a very real danger @anthonydillon. I'm not sure we should really have a fast phase.

deadlight:
We need some exceptions: for instance, in my HTML doc work I want the PR for creating the file to be landed before I put another in for adding something else about templates. I don't think waiting 24hrs wait will do anything there other than discouraging authors from bothering.

nottrobin:
I think the main problem is, as @deadlight said, that we need to be able to do work building on an existing document, which may not exist yet.

In which case we could define a process for creating new documents whereby there's always a PR to create the doc simply with a title and nothing else, like this: https://github.com/ubuntudesign/webteam-practices/pull/10. These could then have different rules - can be approved and merged immediately by a single team-member.

That would mean we can then continue to work on submitting things into that new document, regardless of what other work is coming in. Each heading in that document can be considered separately.

deadlight:
Good point @nottrobin, that would simplify matters. In this one-off instance, can I consider #11 cleared for merging?

nottrobin:
@deadlight I think so. @anthonydillon, @WillMoggridge ?

anthonydillon:
I am happy with the concept of genesis (fast) reviews and structural and amendments being (slow)

WillMoggridge:
Yeah, all good

Lukewh:
It's probably still good to have at least 1 review of a PR even for pages with a title to have some kind of checkpoint. Just my 2c.

anthonydillon:
@Lukewh agree, all PRs into this repo should have 2 +1's but "slow" PRs need to be left for 24 hours so everyone catches it.

Lukewh:
@anthonydillon 👍

nottrobin:
Agreed @Lukewh, @anthonydillon. I'll codify that in a new PR soon.

anthonydillon:
@nottrobin could you land or close this one down please.

PR #14, opened 2017-01-27
add a couple of HTML rules about img tags to the code standards

deadlight:
### Done
I've added two pretty universal rules about  tags that aren't formally documented anywhere I could find

### DIFF
--- a/coding/html.md
+++ b/coding/html.md
@@ -9,3 +9,7 @@ General rules for all HTML documents are:
  - Tags and attributes should be lowercase (`<p class="intro">` not `<P CLASS="intro"`)
  - Use double quotes for attribute values
  - Close all elements - either a closeing tag (`<p>...</p>`) or self-closing (`<img src="jeff.jpg" />`)
+
+## Images
+  - Use alt attributes on img elements
+  - Use null alt text and no title attribute on img elements for images that Assistive Technology should ignore


coding/html.md
```diff
@@ -9,3 +9,7 @@ General rules for all HTML documents are:
  - Tags and attributes should be lowercase (`<p class="intro">` not `<P CLASS="intro"`)
  - Use double quotes for attribute values
  - Close all elements - either a closeing tag (`<p>...</p>`) or self-closing (`<img src="jeff.jpg" />`)
+
+## Images
+  - Use alt attributes on img elements
+  - Use null alt text and no title attribute on img elements for images that Assistive Technology should ignore
```
nottrobin: Could you clarify this a bit? What is null alt text? Do we have a reference we could link to about this point?

coding/html.md
```diff
@@ -9,3 +9,7 @@ General rules for all HTML documents are:
  - Tags and attributes should be lowercase (`<p class="intro">` not `<P CLASS="intro"`)
  - Use double quotes for attribute values
  - Close all elements - either a closeing tag (`<p>...</p>`) or self-closing (`<img src="jeff.jpg" />`)
+
+## Images
+  - Use alt attributes on img elements
```
anthonydillon: I would say use descriptive alt text about an image. Describe what the image is showing not what it is or filename. Can you also add some examples of dos and don'ts. Might help demo the differences.

coding/html.md
anthonydillon: Could you write this in clearer English please. Also with an example please.

deadlight:
Can these (entirely valid) suggestions be added later. Isn't incomplete but entirely accurate documentation better than none?

deadlight:
Just leaving this as a note to myself to include: https://www.w3.org/TR/WCAG20-TECHS/H67.html

nottrobin:
@anthonydillon I'm with @deadlight. I think it's more helpful to ask for slight tweaks to what @deadlight's kindly written here, rather than asking him to add significantly more content, example etc.

More content can easily come in the form of subsequent pull requests. I'd rather we encourage small incremental additions.

"alt descriptions should be descriptive" or something can easily be added as a new item later on. The fundamental requirement here is that there *are* alt tags.

PR #15, opened 2017-02-22
Clarify the technical term "Null alt text"

deadlight:
## Done
Added a parenthetical clarification to the term "null alt text" to further clarify the meaning.


anthonydillon:
Looks ok. Again might be nice to add some dos and donts. :+1:

deadlight:
@anthonydillon yeah, I was thinking about making an example for relevant tags. A bad IMG tag with all the rules broken vs. a good one. I'll propose something else soon.

PR #18, opened 2017-03-08
Add introductory section about reviewing pull requests

nottrobin:
I'm making the "approval criteria" a sub-heading under "reviewing pull requests" as I think we may want to say more about the guidelines about how we should do the reviewing.

Although my introduction here doesn't say very much, the key point I want to get across is that we should try not to be harsh or sound dictatorial in our discussion. The one concrete way I think we can do this is by *not* using the "Approve" and "Request changes" features.


deadlight:
All makes sense :+1:

WillMoggridge:
👍

PR #17, opened 2017-03-08
Add a section about how to submit suggestions

nottrobin:
For the CONTRIBUTING.md, we should mention more specifically how to submit suggestions.

Here I'm trying to emphasise:

- PRs should remain small
- We should expect that PRs should generate discussion
- That new documents should be created empty apart from a heading, to help future work go faster (as [previously mentioned](https://github.com/ubuntudesign/webteam-practices/pull/13#issuecomment-275695203)).

PR #19, opened 2017-03-08
Add section: Collaborating on large documents

nottrobin:
This is a new subsection under the "Submitting suggestions" heading suggested in #17.

There have been frustrations about the slow pace of making changes through pull requests, so this is a suggestion for how to ease that frustration.

I hope going forward we might evolve a few different patterns for collaboration that mix the technologies at our disposal. Although ultimately I think it is important to keep a rigorous formal process for approving the final version of any documents.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,16 @@ Anyone is welcome to suggest improvements in one of two ways:
 - Start a discussion by submitting [a new issue](https://github.com/ubuntudesign/webteam-practices/issues/new)
 - Suggest a change by submitting a pull request ([how?](https://help.github.com/articles/about-pull-requests/))
 
+## Submitting suggestions
+
+### Collaborating on large documents
+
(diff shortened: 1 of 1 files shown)


nottrobin:
@anthonydillon do you have any other suggestions for how to make extensive changes or additions easier to do? Do you accept that, in the current model, getting, say, 30 new paragraphs approved is going to be a slow laborious process.

anthonydillon:
I concede is my be slow but so is editing anything official. We should be
able to comment and propose changes quickly as a team.

nottrobin:
@anthonydillon cool. I'm happy to shelve this, but we should hear more opinions first.

anthonydillon:
Of course and this is just my opinion. It would be great to hear what
others think.

anthonydillon:
Arm twisted :+1:

deadlight:
I'll say that when it's easy to do so, we should use GitHub but we can use GDocs if we're collaborating or are finding editing via pull request too burdensome. :+1:

WillMoggridge:
👍

PR #23, opened 2017-05-19
Define ports here (instead of the wiki)

nottrobin:
Each of our projects should have a different local development port.

Previously we've defined ports in our private wiki:
https://wiki.canonical.com/ProductStrategyTeam/WebDevelopment/LocalDevPorts

However, there's no reason for this information to be private, since the projects are public. I propose we keep this information here instead.


WillMoggridge:
Good plan 👍

barrymcgee:
✅

PR #25, opened 2017-05-19
A skeleton of the foosball rules

nottrobin:
Have at it!

### DIFF
--- a/foosball.md
+++ b/foosball.md
@@ -1,0 +1,28 @@
+# The team foosball rules
+
+Outlined below are the official Canonical webteam rules for foosball:
+
+## Doubles rules
+
+**Best of three**: The winning side is the first side to win two games
+
+**Win by two**: Each game must be won by a margin of two goals
+
+**Flip for it**: When playing a third game, a coin-toss will decide who gets to choose side
+
+- *"Tails for Wales"*: For the toss, the side with the most convincing association with Wales is tails
+
+**Scorers switch places**: After each goal, the two players on the side who scored the goal must swap places
+
+- *No other switching*: This is the only situation under which the players are allowed to switch places
+
+**Spinning**: Feel free to spin the players (good luck with that)
+
+**Being 10-balled**: The players on a side who is beaten 10-0 in a game must swap shirts/tops
+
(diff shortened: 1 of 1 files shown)


nottrobin:
@anthonydillon and @Lukewh I refer you to [the contributing document](https://github.com/canonical-webteam/practices/blob/master/CONTRIBUTING.md#reviewing-pull-requests):

> The GitHub reviews features ("approve" and "request changes") should not be used, as all discussion is merely a discussion of opinions.

nottrobin:
@Lukewh:

> Please add something relating to ball entry:
> The striking player on 'the ball side of the table' must deploy the ball.

It this necessary? I deliberately haven't mentioned any of the really obvious rules (e.g. "Score by putting the ball in the goal", "A game is won by reaching 10 points", "One player is goalie/defender, the other is midfield/attack" etc.) as the rules could get really long, and I don't think they're necessary. It doesn't take a genius to work out where the ball goes.

We may however want something like "the ball should be entered in as neutral a fashion as possible" - as in people shouldn't spint the ball to gain advantage. Although, which I don't do this, personally I don't really mind people doing it. I like spinning, haven't experienced it giving a significant advantage.

Lukewh:
@nottrobin 
> It this necessary? I deliberately haven't mentioned any of the really obvious rules
I've had a few games with some 'newbies' and they don't know that's the entry rule. Same can be said about the point flicker. I think it's important for everyone to be aware.

nottrobin:
Every newbie I've played with has worked out where to put the ball within 20 seconds or so, and I don't think what we're creating here is an instruction manual - no-one's actually going to read it before playing. There really is only one person who can put the ball in, so I'd rather go with the MVP.

But let's wait for some other opinions.

nottrobin:
@deadlight, @pmahnke, @grahambancroft, @WillMoggridge could you give opinions on @Lukewh's suggestion?

Lukewh:
I think this page is good to merge. If anything else comes up in future we can update it.

nottrobin:
@Lukewh k thanks merging

PR #57, opened 2017-11-21
Describes some labels; removes “blocked” duplication

matthewpaulthomas:
Adds a “Labels” section alongside the “Pipelines” section, describing a few potentially-misused labels.

Removes the “Blocked” pipeline. It’s redundant for “blocked” to be both a pipeline and a label. It works better as a label, because an issue can understandably be blocked whether it is Untriaged, Triaged, a Current Iteration task, In Progress, or in Review/QA.

Fixes a couple of typos.

### DIFF
--- a/project-management/zenhub.md
+++ b/project-management/zenhub.md
@@ -33,7 +33,7 @@ Current squad boards:
 - There should be one issue per person in the `In Progress` pipeline.
 - Issues should have one assignee at any one time. If an issue requires multiple people at different stages it should be handed over and the assignee changed at the appropriate time. When reassigned, issues should be moved back into the `Current Iteration` pipeline.
 - Issues that are part of an epic should have an estimate based on [the fibonacci estimation guide](https://github.com/canonical-webteam/practices/blob/master/project-management/fibonacci-estimation-guide.md).
-- If for some reason you move to a different issue before completing your current issue, that issue should be moved to the `Current Iteration` pipeline or labelled as `blocked`.
+- If for some reason you move to a different issue before completing your current issue, that issue should be labelled as “blocked”.
 - Issues should have a description if the title isn't obvious.
 - Issues do not need to be part of an epic, but most should.
 
@@ -57,7 +57,7 @@ At the end of an iteration these should be either closed (if all issues are comp
 
 ### Current Iteration
 
-Issues waiting to be picked up in the current iteration. These are either issues that are yet to be started, have been reassigned or are on hold, but not blocked.
+Issues waiting to be picked up in the current iteration. These are either issues that are yet to be started, have been reassigned or are on hold.
 
 ### In Progress
 
@@ -65,7 +65,7 @@ Issues being actively worked on.
 
 Only one issue per person in the squad should be here. As much as we'd like, we can't work on 5 things at once.
 
-If you're waiting for something/ someone to continue work on an issue, move it to the `Blocked` pipeline.
+If you're waiting for something/ someone to continue work on an issue, label it as “blocked”.
 
 If you need to move on to something with a higher priority move any issue you're assigned but not working on back to `Current Iteration`.
 
@@ -76,19 +76,3 @@ Any issue that is deemed code/ UX/ design complete but requires a review or QA.
 ### Closed
 
 Anything that is complete. The final resting place for issues and epics.
-
-## Labels
-
-### blocked
-
-The purpose of marking an issue `blocked` is to alert managers that unblocking it requires their help. Don’t mark it as blocked if, for example, it just needs a meeting to discuss it; that’s something you can arrange yourself.
-
-An issue may be blocked in any unclosed Pipeline: for example, a “Triaged” / “Current Iteration” issue might be blocked if it can’t be started until another issue is closed, or a “Review/QA” issue might be blocked if a required reviewer is on leave. All blocked issues should have a comment explaining why they’re blocked.
-
-### good first issue
-
-Lets potential contributors know that [the issue would be a good one to start with](https://help.github.com/articles/helping-new-contributors-find-your-project-with-labels/). For example, a typo or grammar error.
-
-### help wanted
-
-Alerts contributors that the assigned person (or if there is none, the team as a whole) would prefer outside help in fixing the problem.
--- a/project-management/zenhub.md
+++ b/project-management/zenhub.md
@@ -15,7 +15,7 @@ Current squad boards:
 
 ## Iterations
 
-- Iterations are two weeks. They start on a Monday and finishing on a Friday.
+- Iterations are two weeks. They start on a Monday and finish on a Friday.
 - Each iteration should be a milestone in ZenHub in the format `Iteration [Iteration number] - [endyear]-[endmonth]-[endday]`.
 
  For example `Iteration 9 - 2017-11-17` followed by `Iteration 10 - 2017-12-01`.
@@ -33,7 +33,7 @@ Current squad boards:
 - There should be one issue per person in the `In Progress` pipeline.
 - Issues should have one assignee at any one time. If an issue requires multiple people at different stages it should be handed over and the assignee changed at the appropriate time. When reassigned, issues should be moved back into the `Current Iteration` pipeline.
 - Issues that are part of an epic should have an estimate based on [the fibonacci estimation guide](https://github.com/canonical-webteam/practices/blob/master/project-management/fibonacci-estimation-guide.md).
-- If for some reason you move to a different issue before completing your current issue, that issue should be moved in to either the `Current Iteration` or `Blocked` pipelines.
+- If for some reason you move to a different issue before completing your current issue, that issue should be moved to the `Current Iteration` pipeline or labelled as `blocked`.
 - Issues should have a description if the title isn't obvious.
 - Issues do not need to be part of an epic, but most should.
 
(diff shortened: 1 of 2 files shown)


matthewpaulthomas:
> an issue can understandably be blocked whether it is Untriaged, Triaged, a Current Iteration task, In Progress, or in Review/QA

Coincidentally, an example of this at standup today: [snapcraft.io#140](https://github.com/canonical-websites/snapcraft.io/issues/140) has been blocked because it requires work from the snap store team. Yesterday it was dropped from this iteration. Normally, that would have sent it back to Triaged. But just because it happened to be blocked, it’s been left in the “Blocked” pipeline, amidst four other pipelines that are intended only for cards for this iteration.

More commonly, something is blocked while sitting in Triaged, indicating “we can’t start on this until something external happens”. The higher its priority in Triaged, the more important it is for managers to communicate with the team/teams who can unblock it, so that it can be scheduled.

nottrobin:
I think you have a good point that it would be useful to be able to see blocked status for upcoming work or work that has been moved out of an iteration. Although in the case of https://github.com/canonical-websites/snapcraft.io/issues/140, I really think that should just have been moved back into triaged when it was de-scoped.

The downside, and I'd like to know what others think about this, is that by having a "blocked" column we make it easier to see blocked issues and focus on getting them resolved. Whereas if, say, the "in progress" column is 12 issues long with a "blocked" one at the bottom, we don't have much visibility over that blocked review. Also what happens when an issue that was in progress is now blocked? Should it stay in "in progress"? Or back to "Current iteration"?

One option would be to leave it as I think it currently is - have both a label and a column - That way, in progress work that becomes blocked can go in the blocked column, but we can also flag issues in the "triaged" column as blocked if we need to.

In any case, this change can't be merged into the official practices repo unti it's agreed across the squads, so this will be a while.

project-management/zenhub.md
```diff
@@ -69,14 +69,26 @@ If you're waiting for something/ someone to continue work on an issue, move it t
...
-
 ### Review/ QA
 
 Any issue that is deemed code/ UX/ design complete but requires a review or QA.
 
 ### Closed
 
 Anything that is complete. The final resting place for issues and epics.
+
+## Labels
```
tbille: We made this label description [file](https://github.com/canonical-webteam/practices/blob/master/workflow/labels.md) to describe labels

anthonydillon:
Interesting idea and I agree that a label is far more flexible but it is very useful to touch on each actively blocked card daily in stand ups. As Robin mentioned.

I believe the concept of actively blocked and inactively blocked, is important here. To be clear:
- Actively blocked: work items are items that have been started but became blocked but remain part of the current iteration. 
- Inactively blocked: items are blocked items based on future work. Therefore, not actively apart of the current iteration.

To expand on Robins comment, I would like to suggest adding a blocked label to all blocked work, weither is active or inactive. And, if there is something that is actively blocked it remains in the blocked pipeline also.

Therefore, in Matthews example. The card would be moved back into triaged with the blocked label as it's becomes inactively blocked.

At first glance, this does cause duplication of state and to remedy this maybe the pipeline could be renamed "Actively blocked" but I believe the benefits warrant this symmetry between labels and pipelines.

anthonydillon:
Been 5 days guys, what's the plan for this PR? Do we need a meeting or can we continue to discuss here?

matthewpaulthomas:
> We made this label description file to describe labels

Oh, I didn’t see that, thanks. I’ll move this to the correct file.

> if, say, the "in progress" column is 12 issues long with a "blocked" one at the bottom, we don't have much visibility over that blocked issue - we might forget it needs resolving

> it is very useful to touch on each actively blocked card daily in stand ups.

Snappy squad standups recently changed from reviewing issues person-by-person to reviewing issues pipeline-by-pipeline. With the person-by-person method, I guess you could possibly miss a “blocked” item if it was assigned to someone who was absent that day. Apart from that, though, with either method, we review every “In Progress” and “Review/QA” card every day, so we wouldn’t neglect blocked ones. We don’t review the “Triaged” column, but whether an issue is blocked there matters mainly during scheduling, not during standups.

> Also what happens when an issue that was in progress is now blocked? Should it stay in "in progress"? Or back to "Current iteration"?

It should stay in In Progress, because it’s more important for managers to unblock blocked “In progress” work (since it may involve, for example, partially written code that’s gathering conflicts) than blocked “Current iteration” work (since it hasn’t been started yet).

> I believe the concept of actively blocked and inactively blocked, is important here.

Every bit of metadata in an issue tracker is useful only to the extent that it helps people make better use of their time. In many trackers I’ve noticed a temptation to make (and occasionally erred in making) distinctions that may be highly meaningful, but don’t — or shouldn’t — actually make any difference to what people do with their time. (“Won’t Fix” vs. “Invalid” is a classic example in many trackers.) That’s why in this PR I added not a definition, but a _purpose_ of the “blocked” label: to alert your managers that unblocking it requires their help.

If that purpose is correct, then as in the example above, managers can tell which “blocked” items are most urgent to unblock by which pipeline they are in. Having a “Blocked” pipeline is less helpful for this purpose, not more.

Lukewh:
I'm happy with either approach as long as we're consistent.

I think both labels and pipeline both have value, and perhaps renaming the pipeline to reflect the blocked items are in the current iteration is a good solution. When you start a new iteration the blocked cards (and parent epic) move with you.

anthonydillon:
Poll of options:
Please :+1: the option you agree with.

anthonydillon:
## Remove blocked pipeline

This means deleting the blocked column on the squad boards and use the "Blocked" label to reflect the status of a blocked task.

anthonydillon:
## Use both labels and pipeline

This would leave the blocked column which will house blocked tasks that are part of this iteration. Issues in this iteration would have the blocked label and reside in the blocked pipeline. Tasks that are not apart of this iteration would have the blocked label and reside in the backlog.

pmahnke:
I think I agree with @matthewpaulthomas if an issue is blocked, I would think that it means that someone is waiting for something to happen to unblock it.  So 'forgetting' about it, for now, is ok.  If it is important, or something 'unblocks' it, someone will be pushed/reminded to change the label and lane.

Lukewh:
@anthonydillon in the first option, where does the blocked labelled cards go? Do you keep them 'in progress' if you're actively waiting, do you move them to iteration backlog? What if something is waiting to be QA'd but something came up that then blocks it?

pmahnke:
@Lukewh I assume you would move it to the most appropriate lane... back to triaged if no work can be completed... stay in review if you are waiting on IS...

anthonydillon:
@Lukewh yes it goes to the most suitable lane.

> because an issue can understandably be blocked whether it is Untriaged, Triaged, a Current Iteration task, In Progress, or in Review/QA.

_From @matthewpaulthomas_

kasiag:
I voted for keeping both label and column. It applies to the current lifecycle of issues.

spencerbygraves:
>Snappy squad standups recently changed from reviewing issues person-by-person to reviewing issues pipeline-by-pipeline.

This seems to be the key point. If it's blocked in this iteration you'll see it in stand up, so no need for the blocked lane.

workflow/labels.md
```diff
@@ -37,7 +37,7 @@ Each review label has 3 steps:
 
 | Label   | Description                                                                            |
 | ----    | ---                                                                                    |
-| Blocked | The issue is block by another issue (put it in the description)                        |
+| Blocked | Alerts managers that unblocking it requires their help. Any blocked issue should have a comment explaining why it’s blocked. Don’t mark it as blocked if, for example, it just needs a meeting to discuss it; that’s something you can arrange yourself. An issue may be blocked in any unclosed Pipeline. |
```
nottrobin: This is quite long.

I don't think blocked is always related to managers. Often someone might be blocked waiting for external people without it necessarily needing managers help. For example, we often have things blocked waiting for an RT from IS. This is often not something managers can or should help with.

I also don't think we need "an issue may be blocked in any unclosed Pipeline" - although it's interesting right now as it's different from our current process, it is just the normal way that labels work, so I don't think it needs specifying.

I think the point about arranging your own meetings is just one example of how blockers could be resolved. Maybe we need another section somewhere in our guidelines about ways to resolve common blockers. I'm not sure this table is the place for it.

How about simply:

> Work cannot continue until something is resolved that's outside the assignee's control. The issue should have a comment explaining the blocker.

anthonydillon:
So we have a draw. I would like to suggest, that in situations like this we take the option which involves change (with, of course. The possibility to reverse if it's not working). By giving the change option this slight advantage it can help us from becoming to locally accustom to our process and adverse from changing it.

The option with the change is to merge this PR and delete the "Blocked" pipelines from our squad boards. 

Therefore I am :+1:

nottrobin:
@anthonydillon sounds good to me - is everyone on board with that?

workflow/labels.md
matthewpaulthomas: > For example, we often have things blocked waiting for an RT from IS. This is often not something managers can or should help with.

If you’re confident IS will do the work in time, it isn’t blocked in any sense that others need to know about (though you might mention it at pipeline-based standups). If you aren’t confident, either it should be rescheduled, which implies management approval; or it should be escalated, which also involves management (“[please ask your manager to increase the priority of the request within the RT system](https://wiki.canonical.com/InformationInfrastructure/IS#Escalating_Requests)”).

> I also don't think we need "an issue may be blocked in any unclosed Pipeline"

Fair enough, removed.

> I think the point about arranging your own meetings is just one example of how blockers could be resolved.

I included that only as a common example of something that is not actually blocked. Happy to remove it though.

kasiag:
+1

frankiegranato:
+1

PR #62, opened 2018-04-05
Remove "Type: {Bug,Enhancement,Maintenance}"

nottrobin:
Assigning a "type" to issues, to differentiate between "Bug" and
"Enhancement" takes up a non-zero amount of time. Mostly @shivjoshi1996,
who has to set this label on tons of issues every fricking morning.

I suggest that this classification is useless and has never been used.
Firstly, differentiating between "Bug", "Enhancement" and "Maintenance"
is difficult, and line is blurry. But more importantly, this distinction
has never ever been used, to my knowledge.

I'm frustrated every morning by the management of these useless labels. So let's discuss it - who wants to make a case for the use of these labels?


anthonydillon:
I use them regularly when triaging but I agree they really serve no function. I had dreams of getting analytics/reports based on the "types" of work we were doing. Happy to remove them.

nottrobin:
> @anthonydillon Happy to remove them.

Happy enough to give me a :+1:?

anthonydillon:
Of course :+1:

pmahnke:
@nottrobin they were brought in from Launchpad, I use them, but in a passive agressive way to single that work in an enhancement, not a bug, but I hate being that way. ;-)

+1

PR #70, opened 2018-07-01
Amend progressive enhancement general rule

squidsoup:
While progressive enhancement is possible in an isomorphic react app, this is technically impossible for us given that we are unable to use node serverside. Even without the node restriction, universal react apps are significantly more complex, for potentially little gain.  I don't think this rule is pertinent now.

### DIFF
--- a/coding/js.md
+++ b/coding/js.md
@@ -5,8 +5,9 @@ This document outlines the rules for writing Javascript documents and fragments 
 Js should be written using [AirBnB](https://github.com/airbnb/javascript) style. New projects can adhere to this by selecting 'AirBnB' during `eslint init`, and existing projects can find the relevant eslint config in [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb).
 
 ## Progressive Enhancement
-Javascript should progressively enhance a page. Core content and functionality should still be available without Javascript.
-The exception to this, is in the case of non-isomorphic SPAs where this is not technically feasible.
+JavaScript should progressively enhance a page.
+
+On all publicly accessible websites, core content and functionality should still be available without JavaScript. JavaScript applications written for a more specific audience which won't be indexed by search engines may be more lax, but should still treat progressive enhancement as a guiding principle.
 
 ## Element hooks
 We should target page element in Javascript using the `data-js` element attribute instead of using the class attribute. Classnames should be used for styling only.
--- a/coding/js.md
+++ b/coding/js.md
@@ -2,6 +2,7 @@
 This document outlines the rules for writing Javascript documents and fragments across all of our codebases.
 
 ## General rules
+-- Variable names should be camelCase
 - Avoid abbreviated or deliberately short names (choose `query` over `q`).
 - If comments are required, they should be inline and above the Javascript they relate to.
 - Javascript code should be indented by two spaces.
--- a/coding/js.md
+++ b/coding/js.md
@@ -2,7 +2,8 @@
 This document outlines the rules for writing Javascript documents and fragments across all of our codebases.
 
 ## General rules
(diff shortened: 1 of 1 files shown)


barrymcgee:
I do not think this should be removed, but it should be amended with the caveat you outline above, i.e. - while the core functionality of certain Javascript-driven product apps is difficult to replicate without server-side rendering, we aspire to develop using the principles of progressive enhancement wherever possible.

Also, even where the above is true - `<noscript>` etc should be utilised to ensure no user gets a white, blank because JS excecution has failed.

Afterall, our biggest site (ubuntu.com) and the majority, if not all, of our marketing sites are initially rendered server-side where this principle should continue to absolutely apply.

squidsoup:
Yep quite right Barry, I’ll update the copy, such that it only applies to
SPAs.

nottrobin:
The core content not being available on SPA apps has significant downsides which mustn't be overlooked - in tutorials.ubuntu.com (polymer) we had serious issues with Google not indexing content, which Will managed to solve by rendering some content server-side.

I really don't think we should include wording that goes anywhere near suggesting that SPAs get a free pass on the progressive enhancement rule.

squidsoup:
In the case of a react application, I don’t see how it is technically
possible to provide the user with anything meaningful without server side
react, which as far as I understand, we are forbidden from using.

squidsoup:
Another important distinction is an authenticated application vs. a
publicly available website like tutorials. There are very good reasons to
provide a fallback for the later, less so for the former.

squidsoup:
I think the important distinction to be made here is between public facing, content based websites, and applications where the execution environment happens to be the browser.

MAAS is a good example of the later however, and provides an SPA without progressive enhancement. This category of application is:

1. Authenticated/private
2. Not constrained by low powered clients, or high latency

I don't think there is much to be gained by requiring progressive enhancement for the later category of application. It introduces significant complexity for little gain in that context. I totally agree however that progressive enhancement should be a requirement for websites that need to be accessible to the general public, and indexable by google.

If we do want to pursue progressive enhancement as the standard for all of our applications, regardless of context, we will need to address our position on node on the server, as technically it is not feasible to build an isomorphic react client without it.

nottrobin:
@squidsoup Ah yes, that's a good point. Previously, these practices really only applied to websites, but I agree that now they should be applicable to interfaces for applications running away from the web.

In which case, would you be happy to rewrite your update not to mention SPAs per se, but instead to mention that the progressive enhancement practice only strictly applies to websites on the actual web, and could be relaxed where we can make certain guarantees about the audience - although it should still be carefully considered?

Although, in the case of MAAS, is it really true that it's "Not constrained by low powered clients"? If I was a network administrator I might very well want to view or edit the settings from my phone on my way to the toilet, during a meeting, or in the building's cafe. And similarly, we have no control over which browser the user is choosing to use, which indeed could even be something like lynx, if only it were supported.

nottrobin:
So we could simply leave it as something like:

> Javascript should progressively enhance a page. On all publicly accessible websites, core content and functionality should still be available without Javascript.

Or we could expand it with a second paragraph like:

> JavaScript applications written for a more specific audience and won't be indexed by search engines may be more lax, but should still treat progressive enhancement as a guiding principle.

Thoughts?

anthonydillon:
I think @nottrobin point is valid. The term SPA is one of many and maybe replaced in the future. I prefer the districting to be between public websites and private applications. We shouldn't mandate the exclusion of progressive enhancement but considered per project.

squidsoup:
Thanks @nottrobin, I think your suggested copy reads well, and I've updated the PR.

nottrobin:
:+1: thanks @squidsoup, looks good to me =)

PR #71, opened 2018-07-01
Adopt AirBnB style.

squidsoup:
Rather than just providing a link to the AirBnB style guide, adopt this style officially.


steverydz:
:+1:

anthonydillon:
LGTM :+1: If we decide we disagree with an element of the style we could fork a Canonical version but until then this works for me.

PR #79, opened 2018-08-07
Add patterns for Discourse documentation

nottrobin:


solutions/documentation.md
```diff
@@ -0,0 +1,55 @@
...
+
+We recommend that community documentation content should live in [Discourse forums](https://www.discourse.org/). Discourse forums provide a useful model for building a community around our content.
+
+This content could live inside a category inside [discourse.ubuntu.com](https://discourse.ubuntu.com/) or any other Discourse installation.
+
+#### Structure
+
+Documentation content should be stored in ["wiki" posts](https://meta.discourse.org/t/what-is-a-wiki-post/30801) that users can comment on, and when they build up enough trust, can edit directly. The [trust levels](https://blog.discourse.org/2018/06/understanding-discourse-trust-levels/) have configurable thresholds. Documentation topics should be grouped under an appropriately-named [category](https://meta.discourse.org/t/how-to-add-categories/71859) - e.g. `docs` or `product-docs`.
+
+Each documentation set shuold have an introductory page. At the bottom of this page should be a horizontal rule, followed by the heading "Content", and then a navigation list which should link to other documentation pages in the forum, e.g.:
```
squidsoup: set *should* have (sp)

solutions/documentation.md
```diff
@@ -0,0 +1,55 @@
...
+- [Getting started](/t/getting-started/3876)
+- [Configuration options](/t/configuration-options/87)
+
+```
+
+#### Branching documentation
+
+Every effort should be made to keep one canonical set of documentation. When different versions of software need to be support, any differences should be mentioned inline in the documents.
+
+If branches of documentation must be created, e.g. when a new major version of software, additional categories can be created for the old version - e.g. `docs-v1.x` alongside `docs`. The title of this older content should contain the version number - e.g. `v1.x: Getting started`.
```
squidsoup: instead of "when a new major version of software" perhaps "when a new major version of software is released"?

nottrobin: Oops

solutions/documentation.md
```diff
@@ -0,0 +1,55 @@
...
+Hello. Welcome to our documentation.
+
+Please spend a while here.
+
+---
+
+# Content
+
+- [Getting started](/t/getting-started/3876)
+- [Configuration options](/t/configuration-options/87)
```
anthonydillon: Can we show a small example of one with sectioned navigation? How would that work?

solutions/documentation.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+
+#### Branching documentation
+
+Every effort should be made to keep one canonical set of documentation. When different versions of software need to be support, any differences should be mentioned inline in the documents.
```
anthonydillon: be _supported_, any

solutions/documentation.md
anthonydillon: Maybe remove the major versions comment and replace it with something like: "A major change to the software which results in a large delta between the current docs and the new ones."

solutions/documentation.md
```diff
@@ -0,0 +1,55 @@
...
+
+If branches of documentation must be created, e.g. when a new major version of software, additional categories can be created for the old version - e.g. `docs-v1.x` alongside `docs`. The title of this older content should contain the version number - e.g. `v1.x: Getting started`.
+
+#### Plugins
+
+We discourage the use of plugins in Discourse unless absolutely necessary, as it's easier to manage multiple Discourse installations if they remain as similar as possible.
+
+### The website application
+
+Documentation content will be pulled from the Discourse forum to be displayed on our official documentation websites - e.g. at https://maas.io/docs - which will be written and managed by the Web and Design team.
```
anthonydillon: Instead of showing an example to maas docs. Maybe link to docs.ubuntu.com would be a better canonical link to official documentation websites.

solutions/documentation.md
```diff
@@ -0,0 +1,55 @@
...
+
+We will use a standard Python package, which we maintain, in each of our sites to pull this content and display it. This package will:
+
+- Pull the content live from Discourse (with some caching) so that changes to the content on Discourse will be reflected without significant delay.
+- Pull the "cooked" markup for the documentation page from Discourse, so the Discourse installation remains in control of parsing the markdown content.
```
anthonydillon: Seems you have two points pulling content?

nottrobin: I'll reword

solutions/documentation.md
nottrobin: The application I'm writing isn't opinionated about this. All it will do is convert links so they'll work on the website - other than that, it will simply parse anything below `# Content` faithfully, and output the markup into a `<nav>` section or whatever.

If we are only planning so support certain navigation structures from a styling (Vanilla) perspective, then it would make sense to explain those structures here, but I'm not the right person to write that. So how about I merge this first and then you augment this section with the navigation structures you plan to support?

nottrobin: After discussing with voices, we've agreed I'll add a couple of sections, as this is the structure most documentations sets will have.

nottrobin:
@anthonydillon, @squidsoup I believe I've addressed all your points.

nottrobin:
@squidsoup by the way, the way I originally wrote `CONTRIBUTING.md`, you're [not technically supposed to](https://github.com/canonical-webteam/practices/blob/master/CONTRIBUTING.md#reviewing-pull-requests) officially "Approve" PRs:

> To encourage participation and activity on the project, comments should remain as polite and accommodating as possible. The GitHub reviews features ("approve" and "request changes") should not be used, as all discussion is merely a discussion of opinions.

Although practically everyone has done it at some point, so maybe it was a foolhardy thing to suggest...

solutions/documentation.md
nottrobin: I've said "e.g." because it is only an example, but the most likely one. I don't want to say about a large delta, because I think people are more likely to interpret that as meaning that they can provide a new set of documentation for version `2.5`, which is definitely not recommended.

nottrobin:
@anthonydillon, @squidsoup would either of you be happy enough with my fixes to give me a :+1: =)

anthonydillon:
:+1:

PR #90, opened 2018-10-07
Add code review document

squidsoup:
Add an initial document for code review practices, with some guidelines around etiquette (as discussed in the dev catchup).

I've also added a heading for *Suggestions for Effective Review*, which perhaps we should discuss as a group.

### DIFF
--- a/coding/review.md
+++ b/coding/review.md
@@ -9,6 +9,6 @@
 ### Approval with suggestions for trivial changes
 In some circumstances, if a branch requires a small number of trivial changes you may approve the branch with the caveat that issues you've identified are resolved or considered, rather than marking the branch *needs fixing* (lp) or *request changes* (gh). These changes should be in the category of minor adjustments to file structure, fixing typos, renamings, small css tweaks etc.
 
-If you are the branch author, it is your responsbility to ensure you have addressed these suggestions before merging your branch.
+If you are the branch author, it is your responsibility to ensure you have addressed these suggestions before merging your branch.
 
 Typically if you have co-workers available to you in your timezone, you should seek an additional review if possible.
--- a/coding/review.md
+++ b/coding/review.md
@@ -7,4 +7,4 @@
 ## Etiquette
 Take care not to block landings for branches that only need minor changes.
 
-If a branch requires a small number of changes, favour approving the branch with a cavaet that the issues you've identified are resolved, rather than *needs fixing*[launchpad] or *request changes*[github]. Marking a branch *needs fixing* should be reserved for genuinely broken code, when the author has proposed the wrong solution, or when the code still requires considerable work to be production ready.
+If a branch requires a small number of changes, favour approving the branch with a caveat that the issues you've identified are resolved, rather than *needs fixing*[launchpad] or *request changes*[github]. Marking a branch *needs fixing* should be reserved for genuinely broken code, when the author has proposed the wrong solution, or when the code still requires considerable work to be production ready.
--- a/coding/review.md
+++ b/coding/review.md
@@ -1,0 +1,10 @@
+# Code Review
+
+## Suggestions for Effective Review
+
+[tbd]
+
+## Etiquette
+Take care not to block landings for branches that only need minor changes.
+
+If a branch requires a small number of changes, favour approving the branch with a cavaet that the issues you've identified are resolved, rather than *needs fixing*[launchpad] or *request changes*[github]. Marking a branch *needs fixing* should be reserved for genuinely broken code, when the author has proposed the wrong solution, or when the code still requires considerable work to be production ready.
--- a/coding/review.md
+++ b/coding/review.md
@@ -5,6 +5,8 @@
 [tbd]
 
 ## Etiquette
-Take care not to block landings for branches that only need minor changes.
 
-If a branch requires a small number of changes, favour approving the branch with a caveat that the issues you've identified are resolved, rather than *needs fixing*[launchpad] or *request changes*[github]. Marking a branch *needs fixing* should be reserved for genuinely broken code, when the author has proposed the wrong solution, or when the code still requires considerable work to be production ready.
+### Approval with suggestions for trivial changes
+In some circumstances, if a branch requires a small number of trivial changes you may approve the branch with the caveat that issues you've identified are resolved or considered, rather than marking the branch *needs fixing* (lp) or *request changes* (gh). These changes should be in the category of minor adjustments to file structure, fixing typos, renamings, small css tweaks etc.
+
+Typically if you have co-workers available to you in your timezone, you should seek an additional review if possible.
--- a/coding/review.md
+++ b/coding/review.md
@@ -3,8 +3,8 @@
 ## Etiquette
 
 ### Approval with suggestions for trivial changes
-In some circumstances, if a branch requires a small number of trivial changes you may approve the branch with the caveat that issues you've identified are resolved or considered, rather than marking the branch *needs fixing* (lp) or *request changes* (gh). These changes should be in the category of minor adjustments to file structure, fixing typos, renamings, small css tweaks etc.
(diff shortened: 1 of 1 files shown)


nottrobin:
I'm not sure the "coding" folder is the right place for this. To my mind that's for things that are more specifically about programming.

I think this content is more similar to https://github.com/canonical-webteam/practices/blob/master/content/copy-reviews.md, and perhaps both should be in "workflow/reviews" or something?

Thoughts?

squidsoup:
> However, if a branch requires changes, no matter how small, one should always "request changes" to block merging.

I think this is a bit problematic, particularly for remote devs. In my case, a 'needs fixing' for something trivial, even if it is *required*, means an additional 24 hours to get the branch landed. If the changes are simple, do they really require additional review, or can we trust the author to do the correct thing?

nottrobin:
@squidsoup I think the problem is exactly that we can't necessarily trust the author to do the right thing. If they see a big green "Merge" button, I'd say (from experience) odds are about 50/50 that they'll click it without reading further. And I think even trying to educate the whole team on this will only reduce it to about 25/75.

Personally I'd rather that you have an explicit power to override those sorts of blocking revirws under certain conditions.

nottrobin:
@squidsoup I also have both messed up incredibly small fixes and seen others do so. I also don't agree that the final review should be skipped without good reason.

shivjoshi1996:
@squidsoup can you have another look at this please.

squidsoup:
@shivjoshi1996 Yep, given we couldn't reach consensus I'll close this PR.

squidsoup:
@nottrobin thanks for your thoughts, and appreciate your concern. Perhaps we could discuss this in more detail at the next dev meetup, and then think about what might make sense to have in this doc?

shivjoshi1996:
@nottrobin the message above wasn’t written by me (I’m on holiday), I assume it was by a person running standup (i’ve Edited the original post to show that).

I agree, PR’s in this two shouldn’t be a part of the normal iteration commitments. & should therefore stay open as long as needed.

@squidsoup feel free to reopen if you see fit. 

:)

nottrobin:
We kinda need a stand-up account...

pmahnke:
I think this has been a really busy time and I personally haven't had time to think about it.  We should reopen.

I think the new github changes feature might help with tiny fixes; however, I think skipping a proper review should be a bigger decision as Robin suggests.

Also, since you can base a branch off a branch, can't you nest them and make sure that the first one is reviewed first?

pmahnke:
oh, there is a stand-up account... people just forgot about it

squidsoup:
We broached this topic again in the last developer meeting, and the general consensus seemed to be that it would be helpful to allow for code-review approval with minor changes requested. This would require good judgement on the part of everyone involved and somewhat more responsibility on the part of the branch author.

If we were to consider this, it would be imperative that _only_ the branch author lands their branches after approval. There was some suggestion that perhaps we should generally enforce that anyway.

nottrobin:
> We broached this topic again in the last developer meeting, and the general consensus seemed to be that it would be helpful to allow for code-review approval with minor changes requested. This would require good judgement on the part of everyone involved and somewhat more responsibility on the part of the branch author.

This sounds like a good solution. I think it's basically our soft policy anyway, in that people have been doing this - I've been doing this. So what we're saying here is that the onus is on the reviewer to decide if the changes they've requested are safe to be completed without another review? I'd like it if we could add a note to ask them to err on the side of caution.

> If we were to consider this, it would be imperative that only the branch author lands their branches after approval. There was some suggestion that perhaps we should generally enforce that anyway.

Interesting, has general feeling swung back the other way on this one? I'm happy wither way =)

anthonydillon:
I have stopped merging other people PRs recently and I have not found branches are not landing. Therefore :+1: on switching the policy back.

barrymcgee:
Just to re-iterate the reason we moved to this policy in the first place;

If code has passed all checks and has been peer-reviewed, why should it not be merged asap? By leaving it unmerged, you're just the leaving an extra step for the branch author to complete which you (the reviewer) could complete in a single second when clicking the big green button.

The other big advantage of this approach is that it means `develop` always contained the latest 'approved' version of the codebase so if another person comes along and cuts a new feature branch from develop, there is much less chance of merge conflicts when they come to merge their code from another approved branch which ahs been waiting in the wings for soemthing to press the button.

squidsoup:
> If code has passed all checks and has been peer-reviewed, why should it not be merged asap? By leaving it unmerged, you're just the leaving an extra step for the branch author to complete which you (the reviewer) could complete in a single second when clicking the big green button.

This applies to the case where the branch is approved with minor changes or suggestions. The author needs to be responsible for making those changes as they see fit before merging. If you allow anyone to merge the branch, the author may not have been satisfied with their amendments.

nottrobin:
> This applies to the case where the branch is approved with minor changes or suggestions. The author needs to be responsible for making those changes as they see fit before merging. If you allow anyone to merge the branch, the author may not have been satisfied with their amendments. Only the author is the position to make that judgement, so they need to be solely responsible for merging their branch.

This is a more specific version of the reason I originally pushed for keeping authors as the only people who could merge initially (all that time ago). I often would leave a PR that I thought was finished, and then have [l'esprit de l'escalier](https://en.wikipedia.org/wiki/L%27esprit_de_l%27escalier) moments a bit later on and want to come back and slightly restructure or tidy something up before merging. If I found that PR was merged, I'd probably just never bother making that small improvement.

nottrobin:
@barrymcgee I do think both of the points you raise actually are covered by @anthonydillon's point:

> I have stopped merging other people PRs recently and I have not found branches are not landing. 

Both your points are to do with there being problems arising from an extra delay in merging PRs. I can completely see the points in theory, but I do think whether we ever see them in practice is the real test.

squidsoup:
@nottrobin I've tweaked the copy, could you have another look when you have a moment please?

coding/review.md
```diff
@@ -0,0 +1,14 @@
+# Code Review
+
+## Suggestions for Effective Review
```
nottrobin: I don't think we should merge an empty section. Could you remove this heading and the `[tbd]` for now? Anyone can come along later and add this section if they have time.

coding/review.md
```diff
@@ -0,0 +1,14 @@
...
+
+[tbd]
+
+## Etiquette
+
+### Approval with suggestions for trivial changes
+In some circumstances, if a branch requires a small number of trivial changes you may approve the branch with the caveat that issues you've identified are resolved or considered, rather than marking the branch *needs fixing* (lp) or *request changes* (gh). These changes should be in the category of minor adjustments to file structure, fixing typos, renamings, small css tweaks etc.
```
nottrobin: Would it be okay to expand on `In some circumstances` a bit? What circumstances?

How about:

> Reviewers may consider approving a branch with the caveat that issues they've identified are resolved or considered, rather than marking the branch [*needs fixing* on Launchpad](https://help.launchpad.net/Code/Review#The_flow_of_this_process) or [*request changes* on GitHub](https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/#submitting-your-review).
> 
> They may only do this when a branch requires a small number of trivial changes (e.g. moving files, fixing typos, renaming variables, small CSS tweaks), and they have reason to expect that they may block progress by not being available for a second review – e.g. because of different timezones.
> 
> It is the branch author's responsibility to ensure they have addressed these suggestions before merging their branch. They still should seek an additional review following their changes wherever possible.

squidsoup: Yep, I think that reads well, thanks Robin. Have updated the copy.

nottrobin:
:+1: thanks @squidsoup

steverydz:
:+1:

anthonydillon:
:+1:

nottrobin:
@squidsoup ready to merge whenever you like

PR #92, opened 2018-10-10
Claim port 8030 for docs.snapcraft.io

nottrobin:


local-development/ports.md
```diff
@@ -39,6 +39,7 @@ This is so that team members can easily run multiple projects at the same time w
 | 8027 | [microk8s.io](https://github.com/canonical-websites/microk8s.io)                                  |
 | 8028 | [mir-server.io](https://github.com/canonical-websites/mir-server.io)                              |
 | 8029 | [jaas.ai](https://github.com/canonical-websites/jaas.ai)                                          |
+| 8030 | [docs.snapcraft.io](https://github.com/canonical-websites/docs.snapcraft.io)                      |
```
anthonydillon: Aren't docs in the 82** range?

anthonydillon:
Making -1 to highlight the state of the request.

local-development/ports.md
nottrobin: Well you see, I don't think so. I differentiated the docs with the 82** because the docs repositories didn't live under `canonical-websites`, and were of (I felt) a fundamental different sort of shape.

docs.snapcraft.io now both lives in `canonical-websites` and is a basic Flask app, very similar to our other sites. So I think there's no longer any reason to differentiate it.

anthonydillon: Ok, good reasoning :+1:

PR #98, opened 2018-11-20
Make into a Jekyll site

nottrobin:
Here I've added basic scaffolding to display these practices as a Jekyll site. When this is merged, this should be displayed with GitHub pages at https://canonical-webteam.github.io/practices/.

I copied the markup structure & CSS from https://docs.vanillaframework.io/, which is the best example we have of an official site with a sidebar nav (as far as I'm aware).

I've been quite quick and dirty here, as I don't want to spend too long on this - any solution as all feels like an improvement. So I:

- Simply copied the CSS from that site directly, as I can't be arsed to set up a whole _sass building chain (which is problematic with GitHub pages anyway).
- Lazily used https://assets.ubuntu.com/v1/84489620-picto-knowledge-orange.svg as the logo, and hacked an `<h4>` to give the site a title next to the logo

QA
--

See https://nottrobin.github.io/practices/.

If you want to test it locally, you'll need to `sudo gem install jekyll`, then run `jekyll serve`.


_layouts/base.html
```diff
@@ -0,0 +1,158 @@
...
+    <!-- Google Tag Manager (noscript) --
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ tag_manager_code }}"
+    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+    !-- End Google Tag Manager (noscript) -->
+
+    <div class="docs-container">
+      <header id="navigation" class="p-navigation">
+        <div class="p-navigation__banner">
+          <div class="p-navigation__logo">
+            <a class="p-navigation__link" href="httos://canonical-webteam.github.io/practices">
```
deadlight: You've invented the "httos" protocol here :P

_layouts/base.html
```diff
@@ -0,0 +1,158 @@
...
+              <img class="p-navigation__image" src="https://assets.ubuntu.com/v1/84489620-picto-knowledge-orange.svg" alt="The knowledge owl">
+            </a>
+            <h4 style="padding: 0.5rem 1rem; margin-bottom: 0">Canonical Webteam practices</h4>
```
deadlight: As noted, this is a bit hacky but I think that it's fine

_layouts/base.html
nottrobin: That's not a protocol, it's the HyperText Transfer Obelisk (Secure)

PR #106, opened 2018-11-20
Suggest using create-react-app for bootstrapping

squidsoup:
* New react projects should bootstrap from CRA, rather than using canonical-webteam/react-starter.
* Enumerate preferred libraries.
* Suggest CRBS-UI and canonical-webteam/react-starter as useful references, with caveats.

### DIFF
--- a/coding/react.md
+++ b/coding/react.md
@@ -4,10 +4,10 @@ layout: base
 
 # React coding standards
 
-This document outlines our standards for writing React applications.
+This document outlines our standards for writing React applications and components.
 
-## Starting a new Project
-We recommend [create-react-app](https://github.com/facebook/create-react-app) for bootstrapping a new project, and avoid ejecting as long as feasible.
+## Starting a new React Project
+When starting a new react application, we recommend [create-react-app](https://github.com/facebook/create-react-app) for bootstrapping the project. Avoid [ejecting](https://facebook.github.io/create-react-app/docs/available-scripts#npm-run-eject) as long as feasible to make updating dependencies easier.
 
 ### Preferred libraries
 All projects should generally use the following:
--- a/coding/react.md
+++ b/coding/react.md
@@ -10,11 +10,16 @@ This document outlines our standards for writing React applications.
 We recommend [create-react-app](https://github.com/facebook/create-react-app) for bootstrapping a new project, and avoid ejecting as long as feasible.
 
 ### Preferred libraries
-* routing - [react-router](https://github.com/ReactTraining/react-router)
-* app state management - [redux](https://redux.js.org)
+All projects should generally use the following:
+
 * component testing - [enzyme](https://github.com/airbnb/enzyme)
 * code formatting - [prettier](https://prettier.io) - We realise that not everyone will agree on the choices that Prettier makes, however feel consistency is ultimately more valuable. As has been said in the golang community, "gofmt's style is no one's favorite, yet gofmt is everyone's favorite".
 * linting - [eslint & airbnb](https://www.npmjs.com/package/eslint-config-airbnb)
+
+If you require routing, or state management:
+
+* routing - [react-router](https://github.com/ReactTraining/react-router)
+* app state management - [redux](https://redux.js.org)
 
 If you'd like introduce a new library, or feel we should replace one of the above, please make a PR to start a discussion.
 
--- a/coding/react.md
+++ b/coding/react.md
@@ -12,9 +12,10 @@ When starting a new react application, we recommend [create-react-app](https://g
 ### Preferred libraries
 All projects should generally use the following:
 
+* unittest framework - [jest](https://jestjs.io)
 * component testing - [enzyme](https://github.com/airbnb/enzyme)
-* code formatting - [prettier](https://prettier.io) - We realise that not everyone will agree on the choices that Prettier makes, however feel consistency is ultimately more valuable. As has been said in the golang community, "gofmt's style is no one's favorite, yet gofmt is everyone's favorite".
-* linting - [eslint & airbnb](https://www.npmjs.com/package/eslint-config-airbnb)
+
+Refer to [code-formatting](code-formatting.md) for recommended tools for linting and formatting.
 
 If you require routing, or state management:
 
--- a/coding/react.md
+++ b/coding/react.md
@@ -6,13 +6,23 @@ layout: base
 
 This document outlines our standards for writing React applications.
 
-## Starter Project
+## Starting a new Project
+We recommend [create-react-app](https://github.com/facebook/create-react-app) for bootstrapping a new project, and avoid ejecting as long as feasible.
 
-[canonical-webteam/react-starter](https://github.com/canonical-webteam/react-starter) provides a codified
-example of our standards for react projects (Prettier, AirBnB style).
+### Preferred libraries
+* routing - [react-router](https://github.com/ReactTraining/react-router)
+* app state management - [redux](https://redux.js.org)
+* component testing - [enzyme](https://github.com/airbnb/enzyme)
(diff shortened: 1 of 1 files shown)


coding/react.md
```diff
@@ -6,13 +6,23 @@ layout: base
...
 
-## Starter Project
+## Starting a new Project
+We recommend [create-react-app](https://github.com/facebook/create-react-app) for bootstrapping a new project, and avoid ejecting as long as feasible.
 
-[canonical-webteam/react-starter](https://github.com/canonical-webteam/react-starter) provides a codified
-example of our standards for react projects (Prettier, AirBnB style).
+### Preferred libraries
+* routing - [react-router](https://github.com/ReactTraining/react-router)
+* app state management - [redux](https://redux.js.org)
```
anthonydillon: I don't think we should mandate always using redux? Maybe state that if your application will contain complex state then redux is recommended.

bartaz: I don't read this section as 'mandating'. I guess the intension is just to give us suggestions in terms of: "does you app need state management - consider redux".

Same way not every react app will need routing, not everyone will need redux.

coding/react.md
```diff
@@ -6,13 +6,23 @@ layout: base
...
+* code formatting - [prettier](https://prettier.io) - We realise that not everyone will agree on the choices that Prettier makes, however feel consistency is ultimately more valuable. As has been said in the golang community, "gofmt's style is no one's favorite, yet gofmt is everyone's favorite".
+* linting - [eslint & airbnb](https://www.npmjs.com/package/eslint-config-airbnb)
 
-As much as possible, this project should reflect our standards for React applications. Please make pull requests
-to this project if you feel something should change.
+If you'd like introduce a new library, or feel we should replace one of the above, please make a PR to start a discussion.
+
+### Reference projects
+[CRBS-UI](https://git.launchpad.net/~crbs/crbs/+git/crbs-ui/tree/), while under active development, probably best reflects our standards for react projects currently.
+
```
bartaz: I think it would be good to mention BSI here as well, even if it doesn't fully follow our current starndards and uses older versions of some libs.

```suggestion

[build.snapcraft.io](https://github.com/canonical-websites/build.snapcraft.io) - a bit dated, but first fully React centred project in our webteam. Our first use or React, server-side rendering, Redux, Enzyme, etc. While it doesn't fully follow our current standards it may be a good place to check how some of these libraries were used.

```

coding/react.md
nottrobin: It could be read either way, but I do think there's a subtle thread here that we want to consider.

I noticed that @squidsoup's previous PR on this document changed the wording from "react components" to "react applications". I was gonna comment on it but didn't get around to it before it was merged. The direction we seem to be moving here is towards suggesting that, for projects that use React, React is somehow the *essence* of the application rather than a *tool* used by the application.

For some applications, of course React is very central. But I agree with @anthonydillon that philosophically we should consider React, Redux, the React Router and any other pieces to be tools at our disposal rather than central parts of the framework of our applications. Each should be chosen and used based on its merits.

snapcraft.io is a good example, where we're pulling in pieces of React for certain pages. It definitely shouldn't be considered "a React application".

squidsoup: > The direction we seem to be moving here is towards suggesting that, for projects that use React, React is somehow the essence of the application rather than a tool used by the application.

That sounds perfectly accurate to me. I can't speak for BSI, as I'm not familiar with the codebase, but RBAC is very much a _react_ application and is built using react design patterns (flux). React likes to tout itself as a view library, but it is very much a framework, albeit not a very opinionated one.

squidsoup: @nottrobin I think reading through this diff again, the advice is equally applicable to both react applications (applications like BSI and RBAC that are entirely built in react), and applications that use partially react. The advice around routing and state management naturally just don't apply to the later.

coding/react.md
```diff
@@ -4,15 +4,32 @@ layout: base
...
 
-[canonical-webteam/react-starter](https://github.com/canonical-webteam/react-starter) provides a codified
-example of our standards for react projects (Prettier, AirBnB style).
+### Preferred libraries
+All projects should generally use the following:
 
-As much as possible, this project should reflect our standards for React applications. Please make pull requests
-to this project if you feel something should change.
+* component testing - [enzyme](https://github.com/airbnb/enzyme)
+* code formatting - [prettier](https://prettier.io) - We realise that not everyone will agree on the choices that Prettier makes, however feel consistency is ultimately more valuable. As has been said in the golang community, "gofmt's style is no one's favorite, yet gofmt is everyone's favorite".
```
bartaz: Mentioning code formatting and linting here separately kind of suggests that it may be different from our other standards. React project is a JS project, so it should follow our general JavaScript and code formatting practices. Both prettier and linting are mentioned on [code formatting](https://canonical-webteam.github.io/practices/coding/code-formatting) page.

I don't mind listing these libraries here again (although it is a bit of a duplication that can lead to us having couple of places to update), but it would be good to at least link from here to https://canonical-webteam.github.io/practices/coding/code-formatting and https://canonical-webteam.github.io/practices/coding/javascript

squidsoup: Good suggestion, thanks @bartaz - have removed references to eslint and prettier, and now link to `code-formatting`.

coding/react.md
```diff
@@ -4,15 +4,33 @@ layout: base
...
-This document outlines our standards for writing React applications.
+This document outlines our standards for writing React applications and components.
 
-## Starter Project
+## Starting a new React Project
+When starting a new react application, we recommend [create-react-app](https://github.com/facebook/create-react-app) for bootstrapping the project. Avoid [ejecting](https://facebook.github.io/create-react-app/docs/available-scripts#npm-run-eject) as long as feasible to make updating dependencies easier.
 
-[canonical-webteam/react-starter](https://github.com/canonical-webteam/react-starter) provides a codified
-example of our standards for react projects (Prettier, AirBnB style).
+### Preferred libraries
```
nottrobin: These suggestions are actually general to all JavaScript aren't they? It seems like they'd be better off sitting in their own document about general library choices for JavaScript, or general choices for front-end - similar to https://github.com/canonical-webteam/practices/blob/master/project-structure/server-side-frameworks.md.

Actually, as I'm writing this I realise that actually most of the recommendations in this document should be generalised as JavaScript guidelines - naming of component files and JSX files should probably sit alongside general advice for naming JavaScript files, that advice and advice on Redux could also be applicable in non-React apps. So what should probably happen is that any React-specific advice should form a section in a general document on JavaScript advice.

However, this is a much larger point, beyond the scope of this PR. So ignore it for now =)

I filed https://github.com/canonical-webteam/practices/issues/118 to track this point.

nottrobin:
:+1:

coding/react.md
squidsoup: > These suggestions are actually general to all JavaScript aren't they? 

Of the libraries mentioned, Jest is the only general js lib, the rest are all react specific.

> So what should probably happen is that any React-specific advice should form a section in a general document on JavaScript advice.

Sure, that sounds reasonable. The code formatting doc should probably be merged into the js doc as well.

nottrobin: I thought y'all told me recently that Redux existed independently of React? I can see Enzyme and React-Router only work with React.

squidsoup: > I thought y'all told me recently that Redux existed independently of React? I can see Enzyme and React-Router only work with React.

Sure, but as react is the only contemporary js framework we use in the company, as far as we're concerned `redux` is react specific.

barrymcgee:
👍🏻

coding/react.md
bartaz: Redux is (in theory) independent library. But even their own docs/examples go very quickly from 'using Redux' to 'using Redux with React', that it's quite hard to see them apart, especially that Redux was created with React components in mind.

So as much as it can be used separately, Redux is quite strong part of React ecosystem. But, sure, if we want to have a recommended framework independent state management library - let's make it Redux, so we can use it consistently.

PR #116, opened 2018-11-21
Add active class to current page in sidebar nav

Caleb-Ellis:
## QA

- Pull code
- `jekyll serve`
- Check that links in the sidebar nav correspond to the current page

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -125,6 +125,14 @@
               toggle.classList.toggle('p-icon--close');
               sidebarContent.classList.toggle('u-hide--small');
             });
+
+            // Add active class to sidebar nav
+            var navLinks = document.querySelectorAll('.p-sidebar-nav__item > .p-link--soft');
+            navLinks.forEach(function(link) {
+              if (link.getAttribute("href") === window.location.pathname) {
+                link.className += " is-active";
+              }
+            });
           </script>
         </div>
       </aside>


_layouts/base.html
```diff
@@ -116,14 +116,24 @@ <h4 style="padding: 0.5rem 1rem; margin-bottom: 0">Web and design team practices
             </ul>
           </nav>
           <script>
-            // Toggle mobile sidebar nav
-            var toggle = document.querySelector('.p-sidebar__toggle');
-            var sidebarContent = document.querySelector('.p-sidebar__content');
+            document.addEventListener("DOMContentLoaded", function(event) {
```
nottrobin: I deliberately didn't wrap this in `DOMContentLoaded`, because its placement in the markup is directly after the `<nav>` in question - this leads to the functionality being available literally as soon as possible - as in, as soon as the browser has parsed the `<nav>` section, before the rest of the DOM has loaded.

You should use the `DOMContentLoaded` event iff the `<script>` block was instead included in the `<head>`. Or you could move the script block to the bottom of the document, in which case you'd never need `DOMContentLoaded`. I chose this particular pattern, but I'm happy for you to argue for another. However, if you're going to use `DOMContentLoaded`, this script block should be moved.

Caleb-Ellis: I honestly didn't give it much thought at all, and now that I have it definitely feels unnecessary.

Removed

PR #120, opened 2018-11-22
Adds document for python coding standards

b-m-f:
A small proposal for a document outlining our practices on writing **Python** code.

What do you think about the Testing & Documentation blocks?

### DIFF
--- a/coding/python.md
+++ b/coding/python.md
@@ -1,0 +1,20 @@
+---
+layout: base
+---
+
+# Python code standards
+
+This document outlines the rules for writing Python code across all our Projects.
+
+## Code style and formatting
+
+All Code should be written in accordance with [PEP 8](https://pep8.org/).
+To correctly format all your code please use [black](https://github.com/ambv/black). Since it differs in a small way from the standard **PEP8** make sure to pass `--line-length 79` when using it. The reason for black to do this is outlined [here](https://github.com/ambv/black#line-length).
+
+## Testing
+
+All logic should be tested. If you add something new to the Codebase make sure that you add sufficient testing to ensure longterm maintainability of our projects.
+
+## Documentation
+
+Document your functions with docstrings in accordance with [PEP 257](https://www.python.org/dev/peps/pep-0257/).
--- a/coding/python.md
+++ b/coding/python.md
@@ -8,13 +8,13 @@ This document outlines the rules for writing Python code across all our Projects
 
 ## Code style and formatting
 
-All Code should be written in accordance with [PEP 8](https://pep8.org/).
-To correctly format all your code please use [black](https://github.com/ambv/black). Since it differs in a small way from the standard **PEP8** make sure to pass `--line-length 79` when using it. The reason for black to do this is outlined [here](https://github.com/ambv/black#line-length).
+All code must conform to [PEP 8](https://pep8.org/) and be formatted with [black](https://github.com/ambv/black) using the option `--line-length 79`.
+Any pull requests that do no conform to this should be rejected by the CI.
 
 ## Testing
 
-All logic should be tested. If you add something new to the Codebase make sure that you add sufficient testing to ensure longterm maintainability of our projects.
+Although tests are lacking on many of our existing projects, writing tests is strongly encouraged. Please write tests for any new code if at all possible, and any efforts to improve test coverage on any project would be greatly appreciated.
 
 ## Documentation
 
-Document your functions with docstrings in accordance with [PEP 257](https://www.python.org/dev/peps/pep-0257/).
+You should document your functions with docstrings in accordance with [PEP 257](https://www.python.org/dev/peps/pep-0257/).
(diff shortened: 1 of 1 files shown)


coding/python.md
```diff
@@ -0,0 +1,20 @@
...
+layout: base
+---
+
+# Python code standards
+
+This document outlines the rules for writing Python code across all our Projects.
+
+## Code style and formatting
+
+All Code should be written in accordance with [PEP 8](https://pep8.org/).
```
nottrobin: I think maybe it's worth being clear about which points here are absolute requirements and which are strong suggestions - maybe with a nod towards https://tools.ietf.org/html/rfc2119.

So, in this case I think we can say: 

```suggestion
All code must conform to [PEP 8](https://pep8.org/).
```

coding/python.md
```diff
@@ -0,0 +1,20 @@
...
+To correctly format all your code please use [black](https://github.com/ambv/black). Since it differs in a small way from the standard **PEP8** make sure to pass `--line-length 79` when using it. The reason for black to do this is outlined [here](https://github.com/ambv/black#line-length).
```
nottrobin: I don't think we need to explain Black's choices here - it seems a bit off-topic. People can do that research themselves if they really want to.

We might want to add a note that CI should reject any pull requests that don't conform to these standard though.

coding/python.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Testing
+
+All logic should be tested. If you add something new to the Codebase make sure that you add sufficient testing to ensure longterm maintainability of our projects.
```
nottrobin: Unfortunately I think this is just too far from either the reality of our existing codebases or our current practice to be taken seriously. Maybe we could say something like:

```suggestion
Although tests are lacking on many of our existing projects, writing tests is strongly encouraged. Please write tests for any new code if at all possible, and any efforts to improve test coverage on any project would be greatly appreciated.
```

coding/python.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Documentation
+
+Document your functions with docstrings in accordance with [PEP 257](https://www.python.org/dev/peps/pep-0257/).
```
nottrobin: This is where I think https://tools.ietf.org/html/rfc2119 wording could be useful to make it clear whether this is a hard requirement or not. Again, our current practice falls quite a long way short of writing docstrings for *every* function, and I don't think that it will ever be 100% followed as a hard requirement. So, to be clear, I'd say:

```suggestion
You should document your functions with docstrings in accordance with [PEP 257](https://www.python.org/dev/peps/pep-0257/).
```

coding/python.md
```diff
@@ -0,0 +1,20 @@
...
+---
+
+# Python code standards
+
+This document outlines the rules for writing Python code across all our Projects.
+
+## Code style and formatting
+
+All code must conform to [PEP 8](https://pep8.org/) and be formatted with [black](https://github.com/ambv/black) using the option `--line-length 79`.
+Any pull requests that do no conform to this should be rejected by the CI.
```
nottrobin:
```suggestion

Project should be configured to block merging on pull requests that do not conform to these standards.
```

coding/python.md
```diff
@@ -9,7 +9,7 @@ This document outlines the rules for writing Python code across all our Projects
 ## Code style and formatting
 
 All code must conform to [PEP 8](https://pep8.org/) and be formatted with [black](https://github.com/ambv/black) using the option `--line-length 79`.
-Any pull requests that do no conform to this should be rejected by the CI.
+Project should be configured to block merging on pull requests that do not conform to these standards.
```
nottrobin: I wasn't quite sure if you intended this line to be its own paragraph. If so, it needs a black line above it. But I don't mind either way.

nottrobin:
:+1: thanks for this @b-m-f it's great! Sorry for all the persnickety comments.

Caleb-Ellis:
:+1: from me!

jpmartinspt:
Looking good. Great work @b-m-f

nottrobin:
@b-m-f for future reference, this [shouldn't have been merged](https://github.com/canonical-webteam/practices/blob/master/CONTRIBUTING.md#approval-criteria) for another 18 hours:

> - A reasonable effort should be made to bring it to the attention of the whole team
> - It should also have been +1d by at least 2 team members
> - It should have been open for at least 24 hours

coding/python.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Code style and formatting
+
+All code must conform to [PEP 8](https://pep8.org/) and be formatted with [black](https://github.com/ambv/black) using the option `--line-length 79`.
+
+Project should be configured to block merging on pull requests that do not conform to these standards.
+
+## Testing
+
+Although tests are lacking on many of our existing projects, writing tests is strongly encouraged. Please write tests for any new code if at all possible, and any efforts to improve test coverage on any project would be greatly appreciated.
```
squidsoup: Only strongly encouraged? Shouldn't tests be mandatory?

squidsoup: > And unfortunately we're not in a place at the moment to enforce 100% test coverage on all new work.

Why's that?

nottrobin: I responded in possibly too much depth over at https://github.com/canonical-webteam/practices/issues/122#issuecomment-441273018.

I didn't want to start that discussion as part of this PR because I think it's better to initially document current practice, so we can build up the practices faster, but I'm very happy to explore evolving the policy.

PR #123, opened 2018-11-22
Consolidate react.md and code-formatting.md into javascript.md.

squidsoup:
A little bit of additional advice, but nothing too controversial hopefully.


coding/javascript.md
```diff
@@ -3,33 +3,93 @@ layout: base
...
+### Reference projects
+[CRBS-UI](https://git.launchpad.net/~crbs/crbs/+git/crbs-ui/tree/), while under active development, probably best reflects our standards for react projects currently.
+
+[build.snapcraft.io](https://github.com/canonical-websites/build.snapcraft.io) - a bit dated, but first fully React centred project in the webteam. Our first use of React, server-side rendering, Redux, Enzyme, etc. While it doesn't fully follow our current standards it may be a good place to check how some of these libraries were used.
+
+[canonical-webteam/react-starter](https://github.com/canonical-webteam/react-starter) was created to provide a codified example of our standards for react projects (Prettier, AirBnB style). While potentially still a useful reference, it should mostly be considered deprecated as it may not accurately reflect some of the choices we've made in real-world projects.
+
+### File naming conventions
+
+* Component files should use CamelCase and match the name of the default export (e.g. MyComponent.js).
```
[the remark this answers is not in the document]
hatched: It seems like it's quite common both ways 🤷‍♂️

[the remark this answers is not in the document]
squidsoup: Yep, in this case we're just conforming to the [AirBnB style guide for react](https://github.com/airbnb/javascript/blob/master/react/README.md), but I can see why it might feel a little weird, particularly for a linux user.

coding/stylesheets.md
```diff
@@ -210,3 +210,8 @@ developer working with JavaScript, leave these well alone.
 
 The idea is that -- in order to properly separate our concerns, we
 should never have styling and behaviour bound to the same classes.
+
+
+Code Formatting
```
nottrobin: I'm impressed that you followed the heading style of the existing document here, although it's at odds with other documents, without feeling the needs to do extra work to normalise things. To keep this PR focused on its mission. :heart:

I do think that `stylesheets.md` should probably be restructured a bit to give it a new top-level heading, 'cos at the moment it kinda looks like the document is purely about BEM. But that shouldn't happen in this PR.

coding/javascript.md
```diff
@@ -3,33 +3,93 @@ layout: base
 ---
 
 # JS code standards
+This document outlines the rules for writing Javascript across our codebase.
```
b-m-f: `Javascript` -> `JavaScript` for consistency. Sorry :sweat_smile:

coding/javascript.md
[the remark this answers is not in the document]
steverydz: I think CamelCase is pretty standard for React component files

coding/javascript.md
```diff
@@ -3,33 +3,93 @@ layout: base
...
+If you require routing, or state management:
+
+* routing - [react-router](https://github.com/ReactTraining/react-router)
+* app state management - [redux](https://redux.js.org)
+
+#### Redux specific
+* selector composition and memoisation - [reselect](https://github.com/reduxjs/reselect)
+* async/side-effects management (http, localstorage etc.) - [redux-saga](https://github.com/redux-saga/redux-saga)
+
+If you'd like introduce a new library, or feel we should replace one of the above, please make a PR to start a discussion.
```
steverydz: I think a lot of what redux offers can be achieved through the React context API and the upcoming hooks feature.

squidsoup: > I think a lot of what Redux offers can be achieved through the React context API and the upcoming hooks feature.

Redux provides much more than just improving passing props through deeply nested components - time travel, state debugging via redux tools, and redux middleware (sagas, thunks) for decoupling code with side effects. Hooks are also currently an alpha feature.

hatched: While Redux provides a lot of fancy features, its complexity is overkill for a vast majority of projects. These types of libraries should be chosen on a per-project basis and only after you've exhausted built-in options or simpler projects to backfill a few minor issues. 

With all that said....The line above `All projects should generally use the following:` should probably be removed.

squidsoup: > The line above `All projects should generally use the following:` should probably be removed.

That line refers to vanilla an enzyme, which should be used in all react projects. I thought it was clear that the routing and state management were project dependent?

hatched: I read it as the first statement under the heading for `Preferred Libraries` covering everything following. Probably my fault 😺

PR #130, opened 2018-11-29
Add some initial testing advice for js and react.

squidsoup:


coding/javascript.md
```diff
@@ -37,6 +37,26 @@ We use [yarn](https://yarnpkg.com/en) for dependency management.
...
+
+#### Testing React
+  * Tests should be collocated with their component in the same directory (MyComponent.js should have a corresponding MyComponent.test.js)
+  * If a component seems difficult to test, it might indicate that it needs to be further decomposed into smaller components.
+  * Unittests should test component behaviour, rather than explicit rendering. Typically you want to test state changes,
+    that handlers are called with appropriate arguments, and any internal logic.
+  * [Jest snapshots](https://jestjs.io/docs/en/snapshot-testing) are ideal for catching regressions in component output/rendering. If you are reviewing a branch, do take care to inspect the snapshot diffs.
+  * Favour testing components in isolation (use enzyme's `shallow` over `mount` where possible).
+
+##### Redux
```
huwshimi: I reckon all reducers need a test to ensure they are pure functions as this is where stuff can easily go wrong with Redux, but isn't enforced by the library itself. Might be worth adding a note about that here.

coding/javascript.md
```diff
@@ -37,6 +37,26 @@ We use [yarn](https://yarnpkg.com/en) for dependency management.
 
 We use the [jest](https://jestjs.io) unittest framework.
 
+#### General Testing Principals
+  * Avoid setting up state in `beforeEach`. It should be clear from reading a test in isolation what state it is in.
+  * Use test factories to reduce boilerplate for commonly instantiated objects.
+  * Don't test external code/libraries.
+
+#### Testing React
```
huwshimi: When testing the output of a component we've found it helpful to test small changes (is a string correct, does something have a specific class, does an element exist/not exist in the output) with Enzyme vs doing a whole snapshot. This way it is clear to the reviewer/future coder what is really being tested. Then snapshots can be used for checking the full output where required.

Maybe that's covered in your notes before, but might be helpful to spell out when to use Enzyme and when to use snapshots.

coding/javascript.md
```diff
@@ -37,6 +37,26 @@ We use [Yarn](https://yarnpkg.com/en) for dependency management.
 
 We use the [Jest](https://jestjs.io) unittest framework.
 
+#### General Testing Principals
+  * Avoid setting up state in `beforeEach`. It should be clear from reading a test in isolation what state it is in.
```
nottrobin: Could we define in a little more detail what is meant by "state" here?

coding/javascript.md
```diff
@@ -37,6 +37,26 @@ We use [Yarn](https://yarnpkg.com/en) for dependency management.
...
+  * Use test factories to reduce boilerplate for commonly instantiated objects.
+  * Don't test external code/libraries.
+
+#### Testing React
+  * Tests should be collocated with their component in the same directory (MyComponent.js should have a corresponding MyComponent.test.js)
```
nottrobin: Is this a strong React standard, and not a strong JavaScript one? Because otherwise I think we should choose one structure for all of JavaScript.

squidsoup: There aren't any react standards as such for project structure. This is reasonable advice for any modern js, but given we only use react, I think this is okay to live here?

coding/javascript.md
```diff
@@ -37,6 +37,26 @@ We use [Yarn](https://yarnpkg.com/en) for dependency management.
 
 We use the [Jest](https://jestjs.io) unittest framework.
 
+#### General Testing Principals
+  * Avoid setting up state in `beforeEach`. It should be clear from reading a test in isolation what state it is in.
+  * Use test factories to reduce boilerplate for commonly instantiated objects.
```
bartaz: These first two points seem quite vague without the examples. What "state" we don't want to set in `beforeEach` (and what is OK to be set in there), what are "test factories", what should they produce?

squidsoup: Sure, I'll try to come up with some examples.

nottrobin: I think the "state" point could be a principle, as long as we define what "state" means.

The "factories" point though - this seems an inherently nuanced point about the choices of the individual developer, not significantly distinct from choices of encapsulation in general. Maybe this point should just be removed, until we have enough examples of people "doing it wrong" to distil it into a practice? Or maybe it should be moved to a wiki post or something about tricks that can help with testing. I'm just not sure the use of factories counts as a "principle".

coding/javascript.md
b-m-f: I like this approach alot. Makes it easier to quickly find corresponding tests.

coding/javascript.md
```diff
@@ -91,6 +111,9 @@ If you'd like introduce a new library, or feel we should replace one of the abov
 
 #### Reducers
 
+* Ensure you use [immutable patterns](https://redux.js.org/recipes/structuringreducers/immutableupdatepatterns) for updating the redux store. Using
```
b-m-f: This and the item below are duplicated now :)

squidsoup: thanks!

coding/javascript.md
squidsoup: @nottrobin @bartaz have expanded on this a bit, hope it is clearer now.

coding/javascript.md
```diff
@@ -37,6 +37,28 @@ We use [Yarn](https://yarnpkg.com/en) for dependency management.
 
 We use the [Jest](https://jestjs.io) unittest framework.
 
+#### General Testing Advice
+* Avoid using `beforeEach`. It should be clear from reading a test in isolation what state (e.g. fixtures, mocks) it is in. Using `beforeEach` creates a number of issues:
```
bartaz: Not sure if I agree with this.

Yes, ideally a test function by itself should be self explanatory and could contain both the setup/context of the test and the test itself.
But if there are multiple tests to be run on the same or very similar setup it's worth (and should be encouraged) to remove the duplication and use `beforeEach`.
Sometimes setup may be quite big and complicated (preparing DOM structure, mocking requests, etc). Having this whole code in the test function makes the 'test' part of it harder to read. In build.snapcraft backend tests where we use nock for mocking requests test setup is sometimes much longer then tests themselves (mocking multiple requests to various endpoints and services) - it would be very error prone to have it duplicated in every test just to avoid using `beforeEach`.

I agree that these are valid examples of some issues usage of `beforeEach` may cause and developers writing tests should be aware of them, but having "avoid using `beforeEach`" as very first testing advice seems to strong to me. `beforeEach` is useful and I think it should be used to avoid duplication and to separate test logic from test setup (especially if setup is shared by multiple tests run in the same context).

coding/javascript.md
nottrobin: If it's the advice for JS in general, could we move it into the generic section?

anthonydillon: I don't thing this can be generic to all JS. Not all JS and it's frameworks are written with a component philosophy. This could be worded as the recommended file structure for a project.

coding/javascript.md
squidsoup: @bartaz appreciate the response. How about a compromise where we suggest that tests sharing the *same* setup are within the same `describe` scope, but as soon as the fixtures diverge they should have a new `describe` scope with their own `beforeEach` context?

squidsoup: > But these are all very specific implementation details of tests. I'm not sure how much we need to codify it here as a practice.

I just think it is a generally a problematic code smell, and have worked in code bases where it has lead to very confusing tests, particularly when coupled with testcase inheritance.

bartaz: OK. I guess in our current JS codebase we don't have enough tests and experience in writing tests for our JS code, so first of all we need to start to write more tests and build some practices around that.

> Use `beforeEach` sparingly and with care.

This sounds better than "avoid" (at least more permissive). Maybe later we could find or write more detailed guidelines for writing tests (what to group in `describe`, when it is ok to use `beforeEach` and when to keep setup in test itself, etc) and link from here.

squidsoup: Sorry this branch has been sitting around and getting rather musty. I've updated the advice around `beforeEach` now to be far less dogmatic. Did you want to have another look @bartaz? Thanks!

coding/javascript.md
```diff
@@ -37,6 +37,28 @@ We use [Yarn](https://yarnpkg.com/en) for dependency management.
 
 We use the [Jest](https://jestjs.io) unittest framework.
 
+#### General Testing Advice
+* Take care when using `beforeEach`. It should be clear from reading a test in isolation what state (e.g. fixtures, mocks) it is in. Use your best judgement, but be aware that using `beforeEach` can create a number of issues:
+  * The developer must look in more than one place to understand the test.
+  * As the test suite grows, developers may add fixtures to `beforeEach` that are only used by a subset of tests, and then the onnus is on the developer to understand whether that fixture is relevant for the test they are reading.
+  * A developer may want a fixture in a *similar* state to one already defined in `beforeEach` and chooses to mutate it in their test, reducing clarity. 
+* Don't test external code/libraries.
```
bartaz: It seems like we have very little 'General Testing Advice'.

While "Don't test external code/libraries" is exactly what I would expect to see in this section, first point about `beforeEach` seems quite specific and detailed.

I think that this section would ideally start with and have more points like "Don't test external code/libraries" and maybe later have more implementation detailed advice like the one about `beforeEach`.

One point that we could add here maybe is some advice about what to test? That we should focus on testing public APIs, functions exported by modules, but we don't need to have a unit test of every internal function?
Something like: "Write tests for public APIs of your modules/classes/components. Avoid unit testing internal functions to make tests independent from implementation details."?

coding/javascript.md
```diff
@@ -37,6 +37,28 @@ We use [Yarn](https://yarnpkg.com/en) for dependency management.
...
+
+#### Testing React
+* Tests should be collocated with their component in the same directory (MyComponent.js should have a corresponding MyComponent.test.js)
+* If a component seems difficult to test, it might indicate that it needs to be further decomposed into smaller components.
+* Unittests should test component behaviour, rather than explicit rendering. Typically you want to test state changes,
+    that handlers are called with appropriate arguments, and any internal logic.
+* [Jest snapshots](https://jestjs.io/docs/en/snapshot-testing) are ideal for catching regressions in component output/rendering. If you are reviewing a branch, do take care to inspect the snapshot diffs. Snapshot tests should compliment unit tests rather than replace them, as unit tests better describe the intended behaviour and output of a component.
+* Favour testing components in isolation where possible (use enzyme's `shallow` over `mount` where possible).
```
bartaz: I think we discussed using `react-testing-library` in place of `enzyme`, should we update this point to reflect that?

coding/javascript.md
b-m-f: > One point that we could add here maybe is some advice about what to test? That we should focus on testing public APIs, functions exported by modules, but we don't need to have a unit test of every internal function?
Something like: "Write tests for public APIs of your modules/classes/components. Avoid unit testing internal functions to make tests independent from implementation details."?

Sounds good!

coding/javascript.md
b-m-f: > I think we discussed using `react-testing-library` in place of `enzyme`, should we update this point to reflect that?

This will also affect:
> Unittests should test component behaviour, rather than explicit rendering. Typically you want to test state changes,
    that handlers are called with appropriate arguments, and any internal logic.

Since in `react-testing-library` the testing principle is to observe the output into the DOM rather than any internal logic.
UnitTests would then be appropriate for helper functions etc. Pretty much anything that would live outside of stateful component and could easily be extracted into another module.

squidsoup: > I think we discussed using `react-testing-library` in place of `enzyme`, should we update this point to reflect that?

Probably not, as most of our projects are still using enzyme. I think we discussed trialing `react-testing-library`, but I'm not sure we've yet had a discussion about it in practice.

coding/javascript.md
squidsoup: That does sound good, but how about proposing it in a follow up branch so that we can land something? This doesn't need to be perfect, and we can iterate on it.

squidsoup: > Write tests for public APIs of your modules/classes/components. Avoid unit testing internal functions to make tests independent from implementation details

I don't agree with that advice - unit testing isn't just about correctness, but a tool to help you find the correct design. You should absolutely write tests for internal logic.

b-m-f: Just the question if components are a small enough unit, or actually sth that calls for integration tests.

I have dabbled with internal tests before, since no public APIs are available for internal functions.
But I am happy to approve this, and iterate on it.

Issue #128, opened 2018-11-29
Guidelines for community engagement


nottrobin:
@anthonydillon, @carlaberkers, @clagom, @shivjoshi1996, @pmahnke and I recently discussed the desirability of having some community engagement guidelines in our practices.

I've been working on Discourse recently, and came across this, which I think is a good example to steal ideas from https://github.com/discourse/discourse/blob/master/docs/code-of-conduct.md.

pmahnke:
I haven't really looked into this too much, but we should check out the existing ubuntu CoC and make sure legal looks at ours - https://www.ubuntu.com/legal/contributors/agreement

nottrobin:
Also, as @pmahnke pointed out in person, the Discourse code of conduct is based off http://contributor-covenant.org/version/1/4/ - which we might want to consider as well.

PR #139, opened 2018-12-07
Add some general guidelines around PR etiquette

barrymcgee:


bartaz:
Good guidelines 👍

nottrobin:
These look good :+1:. This topic overlaps slightly with the code of conduct stuff mentioned in https://github.com/canonical-webteam/practices/issues/128. When that document exists, it would be good to link to it from this one, and we should obviously make sure the advice doesn't conflict. But that's for the future. :+1: for now.

workflow/qa-steps.md
```diff
@@ -5,6 +5,17 @@ description: Recommendations for common QA steps when reviewing PRs
 
 In this section you will find a list of steps to take when reviewing PRs, in order to catch unforeseen regressions.
 
+## General guidelines for quality reviews
+
+- Assume positive intent, we all have collective responsibility for the code we write - be nice.
+- Avoid expressing subjective opinions without reasoning. If you think something is bad, you need to explain _why_ you think it is bad.
+- Try not to increase the scope of a PR unnecessarily. If a reviewing a PR leads you to unrelated code smells, create a new issue.
```
nottrobin:
```suggestion
- Try not to increase the scope of a PR unnecessarily. If reviewing a PR leads you to unrelated code smells, create a new issue.
```

workflow/qa-steps.md
```diff
@@ -5,6 +5,17 @@ description: Recommendations for common QA steps when reviewing PRs
 
 In this section you will find a list of steps to take when reviewing PRs, in order to catch unforeseen regressions.
 
+## General guidelines for quality reviews
+
+- Assume positive intent, we all have collective responsibility for the code we write - be nice.
+- Avoid expressing subjective opinions without reasoning. If you think something is bad, you need to explain _why_ you think it is bad.
```
nottrobin: Does _underscore_ do the same thing as *star*? I didn't know that :smile:

workflow/qa-steps.md
```diff
@@ -5,6 +5,17 @@ description: Recommendations for common QA steps when reviewing PRs
...
+- Try not to increase the scope of a PR unnecessarily. If a reviewing a PR leads you to unrelated code smells, create a new issue.
+- Embrace the 'social coding' aspect of Github. If you spot a typo, rather than type a comment explaining where you found the typo and what the correction should be, [suggest the actual code change](https://help.github.com/articles/incorporating-feedback-in-your-pull-request/#applying-a-suggested-change).
+- As above, you can also push refactors and larger code changes back to a PR which the original author can then review.
+- If you engage in a code review, you now have a responsibility to work with the author to get it over the line. Keep an eye out for your feedback being actioned and schedule a re-review in a timely manner.
+- Have an awareness of team cadence. Software Engineering is an endless craft and we can all craft endlessly. Don't let 'perfect' become the enemy of 'good'.
```
nottrobin: Is there an article or something that could illuminate this point a little better? As it is, I'm worried it may be a little too fuzzy. Does this mean I shouldn't point out issues where I see them? Where is the line?

barrymcgee:
@nottrobin @bartaz I added an extra point after your reviews which is the only point I think might be contentious... or perhaps not? Either way, have a 👀

workflow/qa-steps.md
```diff
@@ -5,6 +5,17 @@ description: Recommendations for common QA steps when reviewing PRs
...
 In this section you will find a list of steps to take when reviewing PRs, in order to catch unforeseen regressions.
 
+## General guidelines for quality reviews
+
+- Assume positive intent, we all have collective responsibility for the code we write - be nice.
+- Avoid expressing subjective opinions without reasoning. If you think something is bad, you need to explain _why_ you think it is bad.
+- Try not to increase the scope of a PR unnecessarily. If a reviewing a PR leads you to unrelated code smells, create a new issue.
+- Embrace the 'social coding' aspect of Github. If you spot a typo, rather than type a comment explaining where you found the typo and what the correction should be, [suggest the actual code change](https://help.github.com/articles/incorporating-feedback-in-your-pull-request/#applying-a-suggested-change).
+- As above, you can also push refactors and larger code changes back to a PR which the original author can then review.
+- If you engage in a code review, you now have a responsibility to work with the author to get it over the line. Keep an eye out for your feedback being actioned and schedule a re-review in a timely manner.
```
nottrobin: I'm slightly worried about this point - I don't want anything to discourage people from jumping in and collaborating on PRs. If someone wants to jump into a PR to make one comment but that's all they've got time for, that seems valuable.

I'd rather we had powerful processed for pushing forward with PRs without relying on anyone in particular. For example, I tend to "dismiss" "changes requested" reviews once all the points are addressed. I think this is a good pattern. Another would be not to assume that because someone might have assigned themselves to a PR that you can't also look at it.

Or maybe we could formalise what "assigned" means, and say don't assign yourself unless you can commit. What do you think?

workflow/qa-steps.md
barrymcgee: I agree it's a fuzzy point but I'm afraid it can't be anything but. 

The line has already always been there - we all decide at some stage when a review is complete.

Two articles which articulate this abstract point quite well;

[1] https://chris.charabaruk.com/2013/03/good-enough-versus-perfect/
[2] http://www.dudewho.codes/perfect-code/

workflow/qa-steps.md
barrymcgee: I don't disagree with any of the above - my point here is that PRs shouldn't languish because it's waiting for a reviewer to return to it.

workflow/qa-steps.md
barrymcgee: Yaaaas @nottrobin 🙌🏻

workflow/qa-steps.md
barrymcgee: It does - double underscore will also result in __bold__. No idea why they are two ways of doing it. 🤷🏻‍♂️

workflow/qa-steps.md
nottrobin: Great. So how about simply:

```suggestion
- If you engage in a code review, try to keep an eye out for your feedback being actioned so you can respond, to avoid blocking progress.
```

workflow/qa-steps.md
nottrobin: Yes linking to one of those articles may help

I agree it's a good point for an individual developer to consider, but I think individuals would probably come to very different conclusions.

Some people might interpret this as meaning that it's better to be sloppy than spend the time to be more careful. Personally I think we should be striving to be pretty careful with everything we actually produce, but not procrastinate by radically limiting the scope (MVP basically).

At some point, I hope we can produce a set of overall guiding team principles here somewhere. A principle I've thought would be a good fit there would be "we care about the details". To some, that might very often conflict, in practice, with "perfect is the enemy of good".

I agree this is a good philosophy for people to be aware of, but if we're to make it a practice, I think we should make some effort to put a stake in the ground somewhere.

Could we maybe remove this one for now, pending further discussion? Or change it to something like:

```suggestion
- Avoid being overly pedantic. Always consider carefully if your point is important enough to hold up work being merged.
```

barrymcgee: As a perfect illustration of this ethos in action, I shall accept your compromise suggestion. 😁

nottrobin:
Thanks @barrymcgee for this great work. This sort of content is exactly the sort of thing I was hoping practices would be about - deep statements about attitude that can help to define our team culture.

I'm very happy with it in it's current form. Big :+1: from me. Thanks again!

PR #153, opened 2018-12-20
Fix typo deployement

lzaaldian:
Hello, I found a typo on Workflow page in Deploy section. I think it should be 'deployment' instead of 'deployement'


nottrobin:
Thanks!

Issue #148, opened 2018-12-13
Define what the practices are trying to achieve


nottrobin:
I'm super happy with how the amount of content in this practices repo has exploded recently. It's all a great effort. While we were building up content in this repo, I took the attitude that basically the more content the better.

However (as @bartaz helped me realise), we're now at a place where we have a bunch of different sorts of things in practices:

1. policies - [browser support](https://canonical-webteam.github.io/practices/coding/browser-support.html)
2. explicit rules - [foosball](https://canonical-webteam.github.io/practices/miscellaneous/foosball.html)
3. statements about our coding standards - [python](https://canonical-webteam.github.io/practices/coding/python.html, [javascript](https://canonical-webteam.github.io/practices/coding/javascript.html), [copy-reviews](https://canonical-webteam.github.io/practices/content/copy-reviews.html) (this last one's should be reformatted)
4. recommendations about how we should work - [QA steps](https://canonical-webteam.github.io/practices/workflow/qa-steps.html)
5. more fuzzy philosophical points (like [project philosophy](https://canonical-webteam.github.io/practices/project-structure/project-philosophy.html))
6. project lists - [ports](https://canonical-webteam.github.io/practices/project-structure/ports.html)
7. documentation - [the run script](https://canonical-webteam.github.io/practices/project-structure/the-run-script.html))
8. guides/articles/tutorials - [stylesheets](https://canonical-webteam.github.io/practices/coding/stylesheets.html), [github](https://canonical-webteam.github.io/practices/workflow/github.html), [zenhub](https://canonical-webteam.github.io/practices/project-management/zenhub.html)

Now that I feel the practices are reaching some level of maturity, it might be time to be explicit about what sort of content we expect them to contain. We could create the statement first, as a practice, and then start updating or moving the content to fit in with the practice.

So, personally I feel that the practices should contain statements about how we should work (the whole spectrum from hard "you must" statements to softer "you should" through to "you may want to think about"), but within that, to try to be concise and to the point. We should write the minimum needed to convey our position on a particular topic.

This means that, for me, anything that feels like a "how to" guide, or a tutorial should live outside the practices, but potentially be linked to from within the practices. If the content already exists out on the wider web, we should simply link to it there (no point duplicating - e.g. BEM). If not, I propose that we keep such guides in [this project's wiki](https://github.com/canonical-webteam/practices/wiki), or, even better, turn them into our own blog posts.

[...]

PR #165, opened 2019-02-04
Add guidance on Performance

barrymcgee:
## Done

- Added some guidelines around web page performance

## QA

- Sanity check Performance guidelines
- Add points I missed or suggest corrections as you see fit 🤓

### DIFF
--- a/coding/performance.md
+++ b/coding/performance.md
@@ -36,10 +36,6 @@ How a web page renders has a huge effect on the perception of how responsive tha
 - Limit DOM nodes to a minimum
 - Consider inlining critical CSS
 
-### Reduce HTTP requests
-
-Round trips to the server are expensive. We should be aware that our users may lose connection before they have downloaded all the assets required to complete page load. Reduce HTTP requests as much as possible by concatenating CSS & JS and embedding images and icons using SVG or Base64 formats.
-
 ### Leverage browser caching
 
 As mentioned above, round trips to the server are expensive. Subsequent requests for static assets that do not change frequently can be avoided by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
--- a/coding/performance.md
+++ b/coding/performance.md
@@ -42,7 +42,7 @@ Round trips to the server are expensive. We should be aware that our users may l
 
 ### Leverage browser caching
 
-As mentioned above, round trips to the server are expensive. These requests can be avoided entirely for static assets that do not change frequently by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
+As mentioned above, round trips to the server are expensive. Subsequent requests for static assets that do not change frequently can be avoided by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
 
 ## Testing tools
 
--- a/coding/performance.md
+++ b/coding/performance.md
@@ -44,4 +44,3 @@ As mentioned above, round trips to the server are expensive. Subsequent requests
 
 - [Google Lighthouse](https://developers.google.com/web/tools/lighthouse/)
 - [Webpage Test](https://www.webpagetest.org/)
-
--- a/coding/performance.md
+++ b/coding/performance.md
@@ -1,0 +1,51 @@
+---
+title: Performance
+description: Ensuring what we build is highly performant
+---
+
+We have a responsibility to ensure that the products and services we provide on the web perform well.
+
+Many of our users consume our products in parts of the world where connectivity is poor, latency is high and data is expensive.
+
+We should treat performance as design - it's not an optional enhancement.
+
+## Performance considerations
+
+### Use images responsibly
+
+Images are one of the leading causes of bloated web pages. Techniques we should consider to mitigate their cost include;
+
+- Compress and optimise all images before reaching production. Optimising jpegs can remove up to 75% of their original file size.
+- Serve appropriately sized images proportionate to where the browser will display them.
+- Use appropriate file types. If possible, use WebP - a modern image format that provides superior lossless and lossy compression for images on the web.
+- Use SVGs where possible for vector based images.
+- Lazy load images that are not visible on initial page load.
+
+### Third party code
+
+Third party code should only be added after a careful cost benefit analysis. If you are including an externally linked library, you should be extra careful as the size of that library is outside our control.
+
+### Rendering
+
+How a web page renders has a huge effect on the perception of how responsive that page is - even small delays of milliseconds can degrade the user experience.
+
+- Wherever possible, defer the loading of third-party code until after `window.onload`
+- Where the above is not possible, use the `async` attribute on the `<script>` element to prevent blocking.
+- Animations and transforms should be done using CSS instead of Javascript.
+- Don't hijack native browser scroll
+- Limit DOM nodes to a minimum
+- Consider inlining critical CSS
+
+### Reduce HTTP requests
+
+Round trips to the server are expensive. We should be aware that our users may lose connection before they have downloaded all the assets required to complete page load. Reduce HTTP requests as much as possible by concatenating CSS & JS and embedding images and icons using SVG or Base64 formats.
+
+### Leverage browser caching
+
+As mentioned above, round trips to the server are expensive. These requests can be avoided entirely for static assets that do not change frequently by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
+
+## Testing tools
+
+- [Google Lighthouse](https://developers.google.com/web/tools/lighthouse/)
+- [Webpage Test](https://www.webpagetest.org/)
+


coding/performance.md
```diff
@@ -0,0 +1,51 @@
...
+- Limit DOM nodes to a minimum
+- Consider inlining critical CSS
+
+### Reduce HTTP requests
+
+Round trips to the server are expensive. We should be aware that our users may lose connection before they have downloaded all the assets required to complete page load. Reduce HTTP requests as much as possible by concatenating CSS & JS and embedding images and icons using SVG or Base64 formats.
+
+### Leverage browser caching
+
+As mentioned above, round trips to the server are expensive. These requests can be avoided entirely for static assets that do not change frequently by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
```
nottrobin: I feel that "These requests can be avoided entirely for static assets" is a misleading statement. Bounce rates on websites tend to at best be close to 50%, often much higher. This means that at least half our visitors get no benefit from caching.

It's not that long caches aren't great - they are - but no-one should be using them as an excuse for not optimising for first-time visitors.

 ```suggestion
As mentioned above, round trips to the server are expensive. These can be somewhat mitigate by avoiding subsequent requests for static assets that do not change frequently by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
```

File: coding/performance.md [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+- Wherever possible, defer the loading of third-party code until after `window.onload`
+- Where the above is not possible, use the `async` attribute on the `<script>` element to prevent blocking.
+- Animations and transforms should be done using CSS instead of Javascript.
+- Don't hijack native browser scroll
+- Limit DOM nodes to a minimum
+- Consider inlining critical CSS
+
+### Reduce HTTP requests
+
+Round trips to the server are expensive. We should be aware that our users may lose connection before they have downloaded all the assets required to complete page load. Reduce HTTP requests as much as possible by concatenating CSS & JS and embedding images and icons using SVG or Base64 formats.
```
hatched: This isn't always the best case scenario. Sometimes splitting up the file to take advantage of multiple simultaneous requests is more efficient. Embedding images and icons also reduces the ability for the browser to cache the page making subsequent visits or other page visits slower.

barrymcgee: It was my understanding that this was only a consideration once we start utilising HTTP/2, as per:

https://developers.google.com/web/fundamentals/performance/get-started/httprequests-5

@hatched Are you saying we should remove this advice entirely or reword it?

hatched: I think I'd propose removing it as it's a case-by-case decision.

nottrobin: We are using HTTP/2 for most of our sites - everything except www.ubuntu.com, www.canonical.com, partners.ubuntu.com and assets.ubuntu.com. (Although these properties are very important). But we don't yet have HPush for any of them.

However, HTTP/2 only pushes the needle a bit in terms of the calculus. With HTTP/1 there's still a cutoff of size where assets above that size are better off being sent separately for their cacheabilty, and below, they're better being inlined, as the latency associated with the HTTP handshake is probably greater than the transfer time.

We could probably argue until the cows come home about where to put that threshold - and it very much depends on your connection speed / distance from the server. If we care a lot about those in the third world, we should actually be prioritising splitting & cacheability over inlining more.

But my instincts would be that the line falls somewhere between 100-200KB for HTTP/1, and maybe more like 50-100KB for HTTP/2.

nottrobin:
This is a really great contribution. Do you think you might want to make it a blog post?

I've been thinking about what sort of thing should go into practices. Ideally I'd like practices to be really concise, deep statements, that are specific to how we want to work in our team. I think it would be best not to repeat existing best practice, but rather link to it. This is so Practices can be useful, easily consumable, and our stronger statements about genuine team positions won't be diluted by pages of general advice.

I've been exploring this idea here: https://github.com/canonical-webteam/practices/issues/148

So I think general best-practice advice on things you should generally think about when optimising for web performance should maybe not be in practices. A list of well known performance techniques doesn't need replicating in our standards.

I think your short introduction is closer to what I'm thinking about as practices - "We should treat performance as design - it's not an optional enhancement" is maybe still a little fuzzy, but it does allude to a specific team attitude to performance, which is gold. But then it should simply link to a list of "techniques we should consider to mitigate their cost include".

I do think what you've written here is really great. Under my suggested model, I'd suggest this sort of content would be better either published as a blog post (like how I did with https://blog.ubuntu.com/2018/12/12/tricks-for-keeping-a-tidy-git-commit-history) or simply as a [Wiki](https://github.com/canonical-webteam/practices/wiki), and then be linked to from a shorter, more declarative set of statements about performance.

But parts of your advice that could be contextualised to provide specific advice for our team and our sites might be useful. For example, we can do better than mentioning that WebP is a good format. We know WebP isn't supported in IE or Safari, but that we [do support](https://canonical-webteam.github.io/practices/coding/browser-support.html) those browsers. And so we could make a strong statement about what we do about this problem - do we not use WebP? Or do we have a specific technique or set of techniques that we recommend? E.g. Cloudinary.

Similarly, I think many of these statements could be useful as practices if they could be changed (and agreed) from fuzzy statements about "you should maybe think about this" into strong statements that genuinely denote the position of the team, as separate from general best practice. For example, if we had specific team-agreed red lines about third party scripts, or even specific third party scripts that we accept, or prohibit.

Of course, my desire & interpretation of what Practices should be is still only an idea. So if people disagree with me about the nature of Practices, I'm happy to be overruled.

nottrobin:
If you agree with me, then a quick way forward could be simply to move your whole "## Performance considerations" section to [the Wiki](https://github.com/canonical-webteam/practices/wiki), and leave this document as just the introduction, but with a link to that Wiki post.

barrymcgee:
@nottrobin I don't disagree _per se_ but it sounds like you'd prefer if this repo was a set of principles rather than a set of practices?

I didn't even know that wiki existed - the trouble with scattering information like this over several locations is people invariably forget where all the disparate resources are.

I'd rather save the blog post for when I've made some tangible performance improvements one of our sites and then write about how that was achieved.

nottrobin:
> it sounds like you'd prefer if this repo was a set of principles rather than a set of practices?

I want it to be both, I think - it's both statements about how we work in practice (e.g. ["we use Prettier"](https://canonical-webteam.github.io/practices/coding/javascript.html#code-style--formatting)) and guiding principles to help our team make decisions (e.g. ["We aim to support and grow the communities that surround our work."](https://canonical-webteam.github.io/practices/project-structure/project-philosophy.html)).

But in both cases, these statements are opinionated - they specifically state a choice we make as a team that is different than a choice another team might make. They aren't general advice for how all software developers should work, or simply reiterations of existing advice that exists elsewhere.

We do, unfortunately, currently have [replicated general advice](https://canonical-webteam.github.io/practices/coding/stylesheets.html), but I'm suggesting we should stop including that sort of content in practices, and keep it elsewhere instead.

> I didn't even know that wiki existed - the trouble with scattering information like this over several locations is people invariably forget where all the disparate resources are.

Yeah I'm suggesting we make that wiki a thing. We have https://wiki.canonical.com/, but that's private, and I'd like there to be a place for wiki posts.

But anyway, if you link to the article from a page in practices, it's just as findable as if they were in the same article. Obviously there is the danger of link rot - but as it's another closely located space that we control, we also have control over the link rot.

> I'd rather save the blog post for when I've made some tangible performance improvements one of our sites and then write about how that was achieved.

I think this would be a great blog post in and of itself - distinctly different from, and maybe more useful, than a post about how we actually improved performance. Although we should write that one too. I really don't want the barrier to publishing blog posts to be high, and this really would be a great addition (maybe with a couple of links to some sources).

nottrobin:
> Yeah I'm suggesting we make that wiki a thing. We have https://wiki.canonical.com/, but that's private, and I'd like there to be a place for publicly available wiki posts.

Or we could try to get a team area on the Ubuntu wiki: https://wiki.ubuntu.com/Teams

nottrobin:
:+1: from me, I guess we should decide the nature of practices separately, and apply it broadly.

coding/performance.md
```diff
@@ -0,0 +1,47 @@
...
+- Consider inlining critical CSS
+
+### Leverage browser caching
+
+As mentioned above, round trips to the server are expensive. Subsequent requests for static assets that do not change frequently can be avoided by specifying an explicit caching policy. However, do not forget to implement a strategy to invalidate the cache if and when those assets do change.
+
+## Testing tools
+
+- [Google Lighthouse](https://developers.google.com/web/tools/lighthouse/)
+- [Webpage Test](https://www.webpagetest.org/)
```
hatched: Last week you were using a couple other tools, maybe those could be added here as well.

barrymcgee: That was Speedcurve which is a subscription SaaS.. however it's built on top of Webpage Test which is already listed.

PR #166, opened 2019-02-04
Small spelling and grammar tweaks

barrymcgee:
## QA

Quick sanity check.

### DIFF
--- a/project-structure/general-structure.md
+++ b/project-structure/general-structure.md
@@ -6,7 +6,7 @@ description: The basic structure for our projects
 Each of our projects follow a similar overall structure, intending to follow [our philosophy for projects](project-philosophy):
 
 - [A `./run` script](https://github.com/canonical-webteam/practices/blob/master/local-development/the-run-script.md) which will install dependencies and run the project using [Docker](https://docs.docker.com/install/) containers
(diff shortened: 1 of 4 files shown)


project-structure/general-structure.md
```diff
@@ -6,7 +6,7 @@ description: The basic structure for our projects
 Each of our projects follow a similar overall structure, intending to follow [our philosophy for projects](project-philosophy):
 
 - [A `./run` script](https://github.com/canonical-webteam/practices/blob/master/local-development/the-run-script.md) which will install dependencies and run the project using [Docker](https://docs.docker.com/install/) containers
-- A `package.json` which defines the basic entrypoint `scripts` to run the project in local development - this is used by the `./run` script, or can be used directly with `yarn run {command}`. It should define the following commands:
+- A `package.json` which defines the basic entry point and `scripts` to run the project in local development - this is used by the `./run` script, or can be used directly with `yarn run {command}`. It should define the following commands:
```
nottrobin: I agree maybe "entrypoint" isn't one word. But I think this was intended to say "entry point `scripts"` - as in the `scripts` are the entry points into the project.

```suggestion
- A `package.json` which defines the basic entry point `scripts` to run the project in local development - this is used by the `./run` script, or can be used directly with `yarn run {command}`. It should define the following commands:
```

PR #167, opened 2019-02-04
Add Prettier ✨

barrymcgee:
## Done

Added Prettier to the codebase for consistency to ensure we continue to eat our own dogfood 🐶

## QA

- Pull branch
- Run `jekyll serve`
- View the site at: http://127.0.0.1:4000/practices/
- Run through all pages and verify they display as expected


package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "devDependencies": {
+    "husky": "^1.3.1",
+    "prettier": "1.16.4",
+    "pretty-quick": "^1.10.0"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged"
```
nottrobin: We discussed previously in dev catch-up that we don't really like the pattern of adding pre-commit hooks to devs' computers without asking.

Could I suggest you just add `prettier`, and then document in the README.md how to use it? If you felt like adding a CI step to check Prettier conformance on PRs, that would be fine too.

barrymcgee: Ok, I've now made that change.

nottrobin: Ah, you simply removed any package.json stuff. Fair enough.

I do think it would be worthwhile adding an `npm run` endpoint, and/or documenting in README.md that you should be linting things with Prettier. But, of course, PRs should remain small, and that stuff can absolutely be added in a future PR.

nottrobin:
LGTM :+1:

PR #168, opened 2019-02-06
Adds troubleshooting sections to README

b-m-f:


README.md
```diff
@@ -33,3 +33,10 @@ jekyll serve
 ## Contributing
 
 For guidelines on contributing to these documents, see [CONTRIBUTING.md](CONTRIBUTING.md).
+
+## Troubleshooting
+
+### jekyll not found
```
barrymcgee:
```suggestion
### Jekyll not found
```

README.md
```diff
@@ -33,3 +33,10 @@ jekyll serve
...
+If you encounter a problem while an installed **gems** that states that it can not be found, make sure that the directory in which **gems** are installed is included in your PATH.
```
barrymcgee:
```suggestion
If you encounter a problem while running `bundle install` that states that Jekyll (or any other gem) cannot be found, make sure that the directory in which **gems** are installed is included in your PATH.
```

b-m-f: This problem should actually be that the `~/.gems/bin` or other dirs were not added to the path. `bundle install` should run correctly when using the specified bundler version :) But I have updated the sentence. 
I hope it is more clear now?

barrymcgee: Cool - thanks, @b-m-f

PR #170, opened 2019-02-07
Add circleci step to run Prettier

barrymcgee:
## Done

As we do not use run Prettier on pre-commit, this change is to run it as a build step instead.

## QA

- Sanity check code
- Check CircleCI Console?

## Question

@nottrobin Do we have a shared config for CircleCI tasks?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - checkout
       - run:
           name: Install dependencies
-          command: bundle install
+          command: yarn install && bundle install
       - run:
           name: Build the site resources
           command: jekyll build
@@ -23,13 +23,13 @@ jobs:
       - run:
           name: Check site is accessible
           command: sleep 1 && curl --head --fail --retry-delay 5 --retry 10  --retry-connrefused http://127.0.0.1:4000/practices/
-  run-prettier:
+  check-prettier:
     <<: *defaults
     steps:
       - checkout
       - run:
           name: Run Prettier over all files
-          command: npx prettier --write **/*.*
+          command: prettier-check **/*
 workflows:
   version: 2
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,38 @@
+---
+defaults: &defaults
+  docker:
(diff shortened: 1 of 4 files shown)


nottrobin:
We don't have a "shared" one, but all our circleci configs do follow a similar pattern:

- They use [our `canonicalwebteam/dev` image](https://hub.docker.com/r/canonicalwebteam/dev/)
- They split up named tasks - e.g. `lint-scss`, `test-python`, `test-site` etc.

[Example](https://github.com/canonical-websites/docs.snapcraft.io/blob/827236f4dd6d935be55da24251dc411f30ada433/.circleci/config.yml)

nottrobin:
So in this case I would suggest just having one job called e.g. "check-prettier", for the time being.

.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+      - run:
+          name: Check site is accessible
+          command: sleep 1 && curl --head --fail --retry-delay 5 --retry 10  --retry-connrefused http://127.0.0.1:4000/practices/
+  run-prettier:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Run Prettier over all files
+          command: npx prettier --write **/*.*
```
nottrobin: I don't think `npx` is available within our development image - could we instead use `node_modules/.bin/prettier`?

Also, I don't think prettier itself is installed, so you probably need a `package.json` with an `Install dependencies` step, with `yarn install`.

Thirdly, won't this command actually do the reformatting? I think what we really want here is just a CI check - as in we probably want `prettier --check`.

package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "devDependencies": {
+    "prettier": "1.16.4",
+    "prettier-check": "2.0.0"
+  },
+  "scripts": {
+    "prettier-check": "prettier-check **/*"
```
nottrobin: Why `prettier-check` instead of simply `prettier --check`? One dependency would be much more preferably to two, if possible. Two opens up the possibility for conflicts and means Renovate has more to do.

PR #174, opened 2019-03-22
Add note about build tools to javascript doc.

squidsoup:
Rollup has generally been considered the standard tool for building libraries in the js community, largely in part to its ability to produce more efficient builds due to [scope hoisting](https://medium.com/adobetech/optimizing-javascript-through-scope-hoisting-2259ef7f5994). 

Webpack however has [supported](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b) scope hoisting via the [ModuleConcatenationPlugin](https://webpack.js.org/plugins/module-concatenation-plugin/) which is the default for production mode builds since v3. This possibly removes one of the more compelling reasons to use Rollup.

The only potential argument for continuing to use Rollup for library builds is that it is generally easier to configure. Whether this incurs additional cognitive overhead, or actually makes things easier for us is probably something the team should discuss. While I'm not opposed to us using both tools, as an experiment, perhaps we should try to configure a similar Webpack build for one of our projects currently using Rollup.

### DIFF
--- a/coding/javascript.md
+++ b/coding/javascript.md
@@ -30,6 +30,7 @@ e.g. `<a data-js="index-link" href="/index">Index</a>`.
 ## Tooling
 
 ### Build tools
+
 [Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) is preferred for libraries.
 
 Rollup and Webpack for the most part have feature parity, so this is mostly a matter of taste and convenience, given Rollup is easier to configure for library builds.
--- a/coding/javascript.md
+++ b/coding/javascript.md
@@ -30,10 +30,9 @@ e.g. `<a data-js="index-link" href="/index">Index</a>`.
 ## Tooling
 
 ### Build tools
-[Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) has generally been preferred for building
-cross environment libraries (both the browser and ESM/CJS environments) if the library in question uses ES6 modules. Rollup and Webpack for the most part have
-feature parity now however (Webpack has support for scope hoisting, and Rollup supports code splitting), so this is mostly a matter of taste and
-convenience, given Rollup is easier to configure for library builds.
+[Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) is preferred for libraries.
+
+Rollup and Webpack for the most part have feature parity (Webpack has support for scope hoisting, and Rollup supports code splitting), so this is mostly a matter of taste and convenience, given Rollup is easier to configure for library builds.
 
 ### Dependency management
 
--- a/coding/javascript.md
+++ b/coding/javascript.md
@@ -32,7 +32,7 @@ e.g. `<a data-js="index-link" href="/index">Index</a>`.
 ### Build tools
 [Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) has generally been preferred for building
 cross environment libraries (both the browser and ESM/CJS environments) if the library in question uses ES6 modules. Rollup and Webpack for the most part have
-feature parity now however (Webpack now has support for scope hoisting, and Rollup now supports code splitting), so this is mostly now a matter of taste and
+feature parity now however (Webpack has support for scope hoisting, and Rollup supports code splitting), so this is mostly a matter of taste and
 convenience, given Rollup is easier to configure for library builds.
 
 ### Dependency management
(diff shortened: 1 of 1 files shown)


coding/javascript.md
```diff
@@ -29,6 +29,12 @@ e.g. `<a data-js="index-link" href="/index">Index</a>`.
 
 ## Tooling
 
+### Build tools
+[Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) has generally been preferred for building
+cross environment libraries (both the browser and ESM/CJS environments) if the library in question uses ES6 modules. Rollup and Webpack for the most part have
+feature parity now however (Webpack now has support for scope hoisting, and Rollup now supports code splitting), so this is mostly now a matter of taste and
```
bartaz: Lots of "now"s in this sentence :)

bartaz:
> The only potential argument for continuing to use Rollup for library builds is that it is generally easier to configure. 
Being a little bit of a devils advocate here: if Rollup is easier to configure and has feature parity with WebPack why do we want to try using WebPack for all projects instead of using Rollup for everything?

We mover from Rollup to WebPack in snapcraft.io some time ago, because once we started having more JS functionality, modules, React and Redux we had build performance issues that made development process harder (slow build, changes to some modules not detected by watch, etc). But maybe it was just the matter of adjusting Rollup config or some tooling that we used for build?

If our goal here is to have potentially one build tool to rule them all this possibly can be webpack. If we can figure out the configuration for the 'library' build for webpack it should be quite straightforward to just reuse it whenever needed without having to learn complicated config settings.

On the other hand I don't really mind having two different tools if one seems to be easier to configure for library builds (rollup) and the other seems to be more suited for application builds (good support for plugins for babel, css modules, react, fast and reliable watch, hot reloading, etc).

squidsoup:
Thanks @bartaz are you +1 then?

bartaz:
Yes, :+1 from me.

Although prettier seems to fail in CI. Not sure if it's related to this PR.
Are we checking .md files for code style?

coding/javascript.md
```diff
@@ -29,6 +29,12 @@ e.g. `<a data-js="index-link" href="/index">Index</a>`.
 
 ## Tooling
 
+### Build tools
+[Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) has generally been preferred for building
```
nottrobin: Would it be okay if we make the language a little more opinionated here, to make our position clearer for people to use?

I'm not quite sure where we landed on the advice - are we really only recommending Rollup for libraries where "cross environment" libraries with "ES6 modules"? It would be much neater if we could simply say - Webpack for applications, Rollup for libraries / packages. Can that be our position?

If so, maybe the language could simply be:

> Webpack is the recommended build tool for applications, Rollup is recommended for packages.
> 
> Rollup and Webpack for the most part have feature parity now however (Webpack has support for scope hoisting, and Rollup supports code splitting), so this is mostly a matter of taste and convenience, given Rollup is easier to configure for library builds.

Thoughts?

nottrobin:
Thanks @squidsoup this is great.

> The only potential argument for continuing to use Rollup for library builds is that it is generally easier to configure. Whether this incurs additional cognitive overhead, or actually makes things easier for us is probably something the team should discuss. While I'm not opposed to us using both tools, as an experiment, perhaps we should try to configure a similar Webpack build for one of our projects currently using Rollup.

If appears that Rollup is our current standard for packages, so let's write it down that way for now, and then we can update it as it evolves.

coding/javascript.md
squidsoup: wfm, thanks @nottrobin

coding/javascript.md
```diff
@@ -29,6 +29,11 @@ e.g. `<a data-js="index-link" href="/index">Index</a>`.
 
 ## Tooling
 
+### Build tools
+[Webpack](https://webpack.js.org/) is the preferred build tool for applications, and [Rollup](https://rollupjs.org/guide/en) is preferred for libraries.
+
+Rollup and Webpack for the most part have feature parity (Webpack has support for scope hoisting, and Rollup supports code splitting), so this is mostly a matter of taste and convenience, given Rollup is easier to configure for library builds.
```
b-m-f:
```
https://github.com/webpack/webpack/tree/master/examples/scope-hoisting
https://webpack.js.org/guides/code-splitting/
https://www.infoq.com/news/2019/03/rollup-code-splitting
https://medium.com/adobetech/optimizing-javascript-through-scope-hoisting-2259ef7f5994#ff06
```
I think we can remove the sentence that one has support and the other does not. They seem to be on par

squidsoup: I think I edited too heavily, it used to read "Webpack *now* has support for scope hoisting"...

PR #179, opened 2019-05-02
Add a supplementary note on code comments

barrymcgee:
Please review for sanity.

### DIFF
--- a/coding/javascript.md
+++ b/coding/javascript.md
@@ -70,6 +70,14 @@ To ensure code formatting consistency across our codebase, we use [Prettier](htt
 
 JavaScript should also be written using [AirBnB](https://github.com/airbnb/javascript) style. New projects can adhere to this by selecting 'AirBnB' during `eslint init`, and existing projects can find the relevant ESLint config in [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb).
 
+### Commenting code
+
+As a general rule, human-readable code is preferred over brevity.
+
+Inline comments should be provided where logic or behaviour is unexpected or required due to external factors.
+
+If library code, [JSDoc](http://usejsdoc.org/) comments should be provided as appropriate.
+
 ## React
 
 ### Starting a new project


coding/javascript.md
```diff
@@ -70,6 +70,14 @@ To ensure code formatting consistency across our codebase, we use [Prettier](htt
 
 JavaScript should also be written using [AirBnB](https://github.com/airbnb/javascript) style. New projects can adhere to this by selecting 'AirBnB' during `eslint init`, and existing projects can find the relevant ESLint config in [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb).
 
+### Commenting code
+
+As a general rule, human-readable code is preferred over brevity.
+
+Inline comments should be provided where logic or behaviour is unexpected or required due to external factors.
+
+If library code, [JSDoc](http://usejsdoc.org/) comments should be provided as appropriate.
```
nottrobin: Is this what we agreed? I don't think I was in dev catch-up for these discussions.

In Python we aim to provide [docstrings](https://www.python.org/dev/peps/pep-0008/#documentation-strings) for all functions (but we don't always succeed). These docstrings don't necessarily need to define parameters or return type or anything, but they should at least contain 1 sentence roughly explaining the function.

Could we recommend a similar thing in JavaScript, as a soft ideal?

coding/javascript.md
```diff
@@ -70,6 +70,14 @@ To ensure code formatting consistency across our codebase, we use [Prettier](htt
 
 JavaScript should also be written using [AirBnB](https://github.com/airbnb/javascript) style. New projects can adhere to this by selecting 'AirBnB' during `eslint init`, and existing projects can find the relevant ESLint config in [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb).
 
+### Commenting code
+
+As a general rule, human-readable code is preferred over brevity.
```
squidsoup: This is a general coding principal, but I'm not sure it belongs under "Commenting code".

coding/javascript.md
```diff
@@ -70,6 +70,14 @@ To ensure code formatting consistency across our codebase, we use [Prettier](htt
...
+
+Inline comments should be provided where logic or behaviour is unexpected or required due to external factors.
```
squidsoup: This is a general principal, not really specific to javascript.

coding/javascript.md
squidsoup: I personally don't think it makes sense to document every function, unless for some reason that function is particularly hard to parse, especially in a react context where the component interface is the most important "documentation" for a consumer.

squidsoup: Maybe distinguish between all library code and public interfaces for library code.

nottrobin: Fair enough. I feel like you've all already discussed this and come to a conclusion in Dev Catch-up, so I'm cool with whatever you agreed. I just wanted to get my point of view on the record =)

nottrobin: As I said I didn't want to throw a spanner in the works if this PR was based on a consensus from the team in a dev catch-up that I missed. However, after speaking to @b-m-f and @jpmartinspt I don't think they felt they'd contributed to a discussion about this either. So I think maybe it is worth opening up this discussion on this PR.

So my position is - as good open-source developers, it is our responsibility to make sure that our code is as easy as possible to pick up and understand by as wide a spectrum of people as possible, with minimal context or understanding needed. Your code *will* be read by others, multiple times, and so the time spent making it easy to understand is never wasted, is extremely valuable. This both helps at every level of working within our team, and also is our responsibility to the open-source community. Not only should any of our devs from any of our squads be able to pick up and understand any of our projects, but ideally a designer, should they feel so inclined, should be able to open up a code file and get a rough idea for what's going on.

Just as we should be doing the heavy-lifting at the software level to create an easy, smooth experience for our end-users, we should be doing the heavy lifting when writing code to make it an easy smooth experience for anyone reading it (the same principle as should apply to writing & communication generally).

And so, I think taking every effort to write clear, simple, "[self-documenting](https://en.wikipedia.org/wiki/Self-documenting_code)" code is an absolute must, and one we should state strongly. However, I don't think writing clean code is ever sufficient. I know there's a tendency in the self-documenting code movement to say that if it needs comments, the code isn't clear enough. I despise this attitude. Nothing is ever going to be as obvious to others as you think it is, and comments are basically always valuable. I have rarely if ever encountered a case where comments have genuinely got in the way of me understanding code (unless they're actual lies). I accept that theoretically, if every line of code had multiple lines of comments it could get pretty dense, but in practice the sparse extra comments I've encountered have never been anything but helpful. 

The point of all this is this - I do think that at least a few words to describe every function would be valuable, as we do in Python. (Of course you could make a distinction about anonymous functions that are directly passed as arguments (`[].forEach(function(){})`).) I personally find JavaScript much more confusing to read than e.g. Python, and so think, if anything, it should have *more* comments. I'd certainly appreciate every named function having a `/** ... */` block comment to describe it, and I'd also appreciate a number of inline comments.

I don't personally see the purpose in making a distinction for "library code", this statement basically explicitly lets you off the hook for commenting non-library code, but I think is a bad idea. I would also really like it if we could consider at least recommending that named functions be described with at least a short sentence.

coding/javascript.md
nottrobin: Yes it is a general principle, but if it applied to Python I think it would be nice if it were a little more encouraging of comments:

``` markdown
Inline comments should be provided in any case where it is not immediately obvious what the code is doing.
```

coding/javascript.md
deadlight: I think that this can be rewritten to get across the idea that if code is so "clever" that it needs comments to document it then the code should be rewritten in an easier to read style.

Lukewh: Perhaps this can be hoisted into the section above `Code style & formatting`?

coding/javascript.md
Lukewh: It is general, but I think it's good to have here too

coding/javascript.md
Lukewh: I read "library code" as "anything that can be consumed in multiple places".
For example in snapcraft we have a [channels.js library](https://github.com/canonical-web-and-design/snapcraft.io/blob/master/static/js/libs/channels.js) that benefits from full jsdocs - it's complex code that does complex things with some seemingly odd rules.
We also have [public/newsletter.js](https://github.com/canonical-web-and-design/snapcraft.io/blob/master/static/js/public/newsletter.js) which is specific, used in one place, and so small that jsdocs really shouldn't be required for anyone to understand what's going on.

In summary I think this sentence is accurate. It doesn't say "if you're writing some complex code that isn't a library DON'T WRITE JSDOC". As with many things in the practices documents, these aren't rules they're guides for people. If you see a PR that would benefit from jsdocs, say.

Lukewh:
It seems this PR has been held up for a long time, I'm :+1: for getting it merged as-is. We don't want stale practices sitting around.

squidsoup:
+1 to merge, the advice here is fine.

PR #192, opened 2019-07-30
Project interface document

nottrobin:
To define the `scripts`, the `Dockerfile` and `./entrypoint`
interfaces for our projects.

### DIFF
--- a/project-structure/standard-interfaces.md
+++ b/project-structure/standard-interfaces.md
@@ -45,9 +45,13 @@ docker build --build-arg BUILD_ID={build-id} .
 
 To enable this, each website projects should contain a `Dockerfile` at the root of the project. This Dockerfile should use [multiple stages](https://docs.docker.com/develop/develop-images/multistage-build/) to do any necessary preparatory steps - e.g. installing dependencies and building assets.
 
-Most images will usually run a service on port `80`, which will typically be run a server through a file called `./entrypoint`, e.g.:
+All images [must be](<https://wiki.canonical.com/InformationInfrastructure/IS/Kubernetes/Policy?highlight=(kubernetes)|(policy)#Docker_images>) built from an official `ubuntu` image. Images should run a service on port `80`, which will typically be started through a file called `./entrypoint`, e.g.:
 
 ```dockerfile
+FROM ubuntu:bionic
+
+# ... other RUN, ADD commands etc.
+
 # Setup commands to run server
 ENTRYPOINT ["./entrypoint"]
 CMD ["0.0.0.0:80"]
--- a/project-structure/standard-interfaces.md
+++ b/project-structure/standard-interfaces.md
@@ -1,0 +1,55 @@
+---
+title: Project interface
+description: The standard interface for our projects - package.json and entrypoint
+---
+
+Our projects should all include [a `package.json` file](https://yarnpkg.com/lang/en/docs/package-json/) at the top-level of the project, which defines the standard ways the project should be run.
+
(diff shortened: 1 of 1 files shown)


project-structure/project-interface.md
```diff
@@ -0,0 +1,45 @@
...
+
+- `start`: The standard entrypoint for local development. It should run any required watchers for rebuilding files as they change, and run any local development servers etc..
+    - E.g.: `concurrently 'yarn run watch' 'yarn run serve'`
+- `build`: Build all the required assets for the project to be published. This may simply run a number of other scripts per filetype, e.g.:
+    - `build-js`: Build the JavaScript files
+    - `build-css`: Build any SCSS files etc. into CSS
+- `serve`: Run the a local webserver. This will typically use the same `./entrypoint` script as is used by the production Docker image (see below).
+- `watch`: Watch for changes in any relevant files. Again, this may be split up into sup-scripts by filetype, e.g.:
+    - `watch-js`: Watch for changes in JavaScript files and rebuild as needed.
+    - `watch-css`: Watch for changes in SCSS files and rebuild CSS as needed.
```
Lukewh: - `test`
  - `test-py`
  - `test-js`
- `lint`
  - `lint-py`
  - `lint-js`
  - `lint-css`
?

nottrobin: I writed some things about this.

anthonydillon:
Nice work, only part missing but not required is a mention of the `.env` file and its usage. :+1:

project-structure/standard-interfaces.md
```diff
@@ -0,0 +1,55 @@
...
+```bash
+yarn run build
+docker build --build-arg TALISKER_REVISION_ID={build-id} .
+```
+
+To enable this, each website projects should contain a `Dockerfile` at the root of the project. This Dockerfile will typically be run a server through a file called `./entrypoint`, e.g.:
+
+```dockerfile
+# Setup commands to run server
+ENTRYPOINT ["./entrypoint"]
```
hatched: It might be nice to add a few other commands here like `RUN` `ADD`  etc, just to show peeps that it's ok to use dockerybits here as well.

nottrobin:
@anthonydillon not mentioning `.env` for now as it's likely going to change.

project-structure/standard-interfaces.md
```diff
@@ -0,0 +1,60 @@
...
+
+The `package.json` file should define some [`scripts`](https://yarnpkg.com/lang/en/docs/package-json/#toc-scripts), and (usually) some [`dependencies` and `devDependencies`](https://yarnpkg.com/lang/en/docs/package-json/#toc-dependencies).
+
+## scripts
+
+These defined `scripts` can be run either with `yarn run {script-name}` or through [our local development tooling](https://canonical-web-and-design.github.io/practices/project-structure/the-run-script.html).
+
+Our standard script names:
+
+- `start` (required): The standard entrypoint for local development. It should run any required watchers for rebuilding files as they change, and run any local development servers etc..
```
bartaz: Do our projects have `start`?  If we treat it as required should we add one to project that don't have it?

Seems like currently the contents of the `scripts` are mostly dictated by run script, so they include `build`, `serve` `watch`, `test` with addition of some more specific ones. 

With this practice, do we want to enforce adding `start` as well? Is `start` supported by run script (like `./run start`)?

nottrobin: Yeah this is the new standard, ad yes we should implement it on projects, and will do when we migrate to the dotrun snap. Maybe you're right that I should have held off on merging until we'd got further in rolling out the new standard.

But that this point I'd rather leave this standard in place and hurry up with my dotrun snap work.

bartaz: Sure, I don't mind. Was just surprised that the only required script is the one I haven't seen in any our project yet.

PR #197, opened 2019-08-13
Update reference react projects.

squidsoup:
Our list of reference react projects was getting a little out of date. I've removed the `react-starter` project completely as it is much more informative looking at any of the other projects.


anthonydillon:
:+1:

barrymcgee:
👍🏻

PR #198, opened 2019-08-13
Suggesting immer for immutable state management.

squidsoup:


coding/javascript.md
```diff
@@ -99,6 +99,7 @@ If you require routing, or state management:
 #### Redux specific
 
 - Selector composition and memoisation - [Reselect](https://github.com/reduxjs/reselect)
+- Immutable state management - [Immer](https://github.com/immerjs/immer)
```
bartaz: Not that I have anything against, but more of curiosity: any reasons about using this particular one?
JS libraries usually tend to have quite short lifespan unless backed by big company or big project.

Redux seems to still suggest good ol' Immutable.js (https://redux.js.org/recipes/using-immutablejs-with-redux)

squidsoup: Immer is **much** simpler than immutablejs, and only touches reducer code. We evaluated immutablejs, but it requires changing the way you write javascript and I gather that it has a tendency to *infect* your codebase, if that makes sense.

PR #200, opened 2019-08-23
Update and add more type labels

pmahnke:
I have added labels that we have everywhere already and three new type ones that I think could enable a rich change log.

| Label            | Description                         |
|------------------|--------------------------------------|
| Bug 🐛           | This is a bug that needs fixing |
| Enhancement ✨   | This is about enhancing functionality |
| Maintenance 🔨   | A required fix to keep the software up to date |


squidsoup:
Excellent. We ended up with some similar ones in `maas-ui`!

image

image

image

image

Not 💯 on "feature" vs "enhancement".

workflow/labels.md
```diff
@@ -43,7 +44,9 @@ Each review label has 3 steps:
 | ------- | ------------------------------------------------------------------------------------------------------------------------------ |
 | Blocked | Alerts managers that unblocking it requires their help. Any blocked issue should have a comment explaining what help you need. |
 | Doing   | Actually working on the issue                                                                                                  |
-| Triaged | A team member has reviewed this issue, agreed that it is valid and assigned a priority                                         |
+| Triaged | A team member has reviewed this issue, agreed that it is valid and assigned a priority                           | Invalid | This isn't a bug. |
```
matthewpaulthomas: - The missing newline here is causing GitHub’s Markdown parser to [omit the “Invalid” row altogether](https://github.com/canonical-web-and-design/practices/blob/pmahnke-patch-1/workflow/labels.md).
- Though it’s changed here accidentally, I’m not aware of any repo that actually has a “Triaged” label. Which is just as well, since the definition of triaged is that it has a priority, which is determined by whether it has a priority label. Labelling an issue as “Triaged” would do nothing to triage it, it would just be noise.
- Currently many/most “Invalid” labels are applied by stalebot, which uses it to mean “inactive”, independent of whether the issue is valid, a bug, both, or neither. If we keep stalebot, we should tweak it so that it uses a different label (like, say, “stale”).

workflow/labels.md
```diff
@@ -43,7 +44,9 @@ Each review label has 3 steps:
...
+| Won't fix | Even if this is a bug, we will not be fixing for a variet or reasons, like it will be replace, it isn't a priority, etc... |
```
matthewpaulthomas: - It’s a constant temptation to classify issues in a way that doesn’t actually improve anyone’s decision-making. I consider it a mistake that I specified “Invalid” and “Won’t Fix” in 2005 for Launchpad’s bug tracker: total time spent by people wondering/arguing/upset about issues’ correct state far outweighed any benefit from distinguishing them. (Of course people can argue about open vs. closed too, but fewer arguable things is better.) The issue trackers in Google Code Project Hosting, then GitHub, were better by having a single closed state.
- If this was retained, `variet or` > `variety of`, `replace` > `replaced`.

workflow/labels.md
```diff
@@ -52,3 +55,11 @@ Each review label has 3 steps:
 | Question         | To ask a question about something on the project                                                                                                                                                                     |
 | Good first issue | Lets potential contributors know that [the issue would be a good one to start with](https://help.github.com/articles/helping-new-contributors-find-your-project-with-labels/). For example, a typo or grammar error. |
 | Help wanted      | Alerts contributors that the assigned person (or if there is none, the team as a whole) would prefer outside help in fixing the problem.                                                                             |
+| Project Fuse ⚡️   | This issue is related to our Project Fuse, which aims to speed up the site for end users |
```
matthewpaulthomas: As I understand it, Project Fuse is specific to public-facing Web sites, but we have plenty of repos that aren’t. How about “performance”?

workflow/labels.md
```diff
@@ -52,3 +55,11 @@ Each review label has 3 steps:
 | Question         | To ask a question about something on the project                                                                                                                                                                     |
 | Good first issue | Lets potential contributors know that [the issue would be a good one to start with](https://help.github.com/articles/helping-new-contributors-find-your-project-with-labels/). For example, a typo or grammar error. |
 | Help wanted      | Alerts contributors that the assigned person (or if there is none, the team as a whole) would prefer outside help in fixing the problem.                                                                             |
+| Project Fuse ⚡️   | This issue is related to our Project Fuse, which aims to speed up the site for end users | 
+| Copy update      | This issue is about a copy update, no design, UX or code changes |
+| Bug 🐛           | This is a bug that needs fixing |
+| Enhancement ✨   | This is about enhancing functionality |
+| Maintenance 🔨   | A required fix to keep the software up to date |
```
matthewpaulthomas: - Many of our repos already have labels “Type: Bug”, “Type: Enhancement”, and “Type: Maintenance”. If these labels were a good idea I think their current names would be better, since they’re grouped in the alphabetical list.
- However, in the repos I’ve seen, these labels are hardly ever used (and I don’t think it’s their length at fault). I think we stopped using them shortly before @nottrobin started the practices repo, on the grounds that using them wouldn’t improve anyone’s decision-making. Sometimes (quite often, actually!) implementing a roadmapped enhancement is higher priority than fixing an obscure bug. And dropping the labels would avoid inconsequential arguments about, for example, whether an accessibility issue counts as a bug or an enhancement.

workflow/labels.md
nottrobin: I certainly agree that people seem to like classify and categorise things which can easily end up causing a lot of busy work, and a slow, subtle drain on people's time.

I certainly agree that the distinction between "invalid" and "won't fix" doesn't seem particularly useful to anyone, and I'd happy drop one or the other. The valuable information here is that "this bug was closed without being implemented", and a single label that communicates that in an inoffensive wording as possible sounds perfectly sufficient to me (for what it's worth, I find both "invalid" and "won't fix" to be somewhat offensive, and would mind if we could find a softer word for the label instead).

workflow/labels.md
```diff
@@ -11,6 +11,7 @@ In this part you will find a description of all the labels we use on issues.
 | ----------- | ---------------------------------------------------------------------------------------- |
 | Don’t merge | While this label is present this pull-request should not be merged                       |
 | Duplicate   | This issue already exists                                                                |
+| Master Epic | This is a master [ZenHub epic](https://www.zenhub.com/blog/working-with-epics-in-github/)|
```
nottrobin: Is it worth explaining what a master epic is a bit more? Or linking to a page that explains the concept?

workflow/labels.md
nottrobin: +1 - It's nice to have codenames for things, but when they end up as labels on public-facing issues, I think that's gone too far and is really hindering understanding.

workflow/labels.md
nottrobin: I agree. I'm unclear whether there's anyone who keenly wants to defend the existence of these distinctions, but if so, please speak up: I'd like to understand what the real-world utility might be to classifying things as "bug"s or "enhancement"s.

workflow/labels.md
```diff
@@ -43,7 +44,9 @@ Each review label has 3 steps:
 | ------- | ------------------------------------------------------------------------------------------------------------------------------ |
 | Blocked | Alerts managers that unblocking it requires their help. Any blocked issue should have a comment explaining what help you need. |
 | Doing   | Actually working on the issue                                                                                                  |
-| Triaged | A team member has reviewed this issue, agreed that it is valid and assigned a priority                                         |
+| Triaged | A team member has reviewed this issue, agreed that it is valid and assigned a priority                           | Invalid | This isn't a bug. |
+| Won't fix | Even if this is a bug, we will not be fixing for a variet or reasons, like it will be replace, it isn't a priority, etc... |
+| Released ✈️ | This issue/PR have been released and live. |
```
nottrobin:
```suggestion
| Released ✈️ | This issue/PR has been released and live. |
```

workflow/labels.md
nottrobin: I agree, my understanding is that "triaged" means "has been assigned a priority", and so that should be indicated by the fact of it having been assigned a priority.

I also think we should strive to use [simple English](https://en.wikipedia.org/wiki/Simple_English_Wikipedia) where possible, so it would be nice if we found other ways to communicate things like "triaged" - and talking in terms of priorities instead is a step in that direction.

workflow/labels.md
matthewpaulthomas: Continuing the Launchpad comparison, in 2006 when I’d realised the error of my ways, I suggested [combining “Invalid” and “Won’t Fix” into “Declined”](https://dev.launchpad.net/IssueTracker#Status) , meaning “Not appropriate or relevant for this project”.

workflow/labels.md
squidsoup: @nottrobin we're using [release drafter](https://github.com/apps/release-drafter) to automatically generate release changelogs from PR commits and labels. Release drafter automatically categorises PRs by label in the draft changelog.

squidsoup: > * Many of our repos already have labels “Type: Bug”, “Type: Enhancement”, and “Type: Maintenance”. If these labels were a good idea I think their current names would be better, since they’re grouped in the alphabetical list.

Prefixing these with "Type:" sounds fine, but we should keep the emojis.

workflow/labels.md
pmahnke: I like 'Declined' I will move to that and remove the others

workflow/labels.md
pmahnke: removing triaged

workflow/labels.md
pmahnke: @squidsoup which ones are important for drafter?

workflow/labels.md
pmahnke: actually, Triaged isn't a label in the webteam, it is a column on zenhub, so let's remove it.

workflow/labels.md
nottrobin: @squidsoup ah right. I didn't realise this was part of that mechanism. That seems like a good argument, for published packages. But it seems kind of a shame to pollute websites (the vast majority of our repos) with those labels without a reason.

squidsoup: > @squidsoup which ones are important for drafter?

Totally customisable, but at the moment we're using "feature", "enhancement", "bug" and "maintenance"

squidsoup: > @squidsoup ah right. I didn't realise this was part of that mechanism. That seems like a good argument, for published packages. But it seems kind of a shame to pollute websites (the vast majority of our repos) with those labels without a reason.

Sure, it's mostly useful if you're using release drafter and github releases - do we keep changelogs for our websites currently?

nottrobin: No we don't have changelogs of websites. Since we're going for as close to a continuous-delivery system as we can, I don't think changelogs make any sense for websites.

nottrobin: Updating following our discussion in dev catch-up - yes releases could be useful for websites if they provide an easy mechanism for us to indicate that an issue / PR had been released. We could look into this.

nottrobin:
These LGTM now :+1:

workflow/labels.md
```diff
@@ -39,16 +40,22 @@ Each review label has 3 steps:
...
+| Question ❓         | To ask a question about something on the project                                                                                                                                                                     |
+| Good first issue 🔰 | Lets potential contributors know that [the issue would be a good one to start with](https://help.github.com/articles/helping-new-contributors-find-your-project-with-labels/). For example, a typo or grammar error. |
+| Help wanted 🏮      | Alerts contributors that the assigned person (or if there is none, the team as a whole) would prefer outside help in fixing the problem.                                                                             |
+| Performance ⚡️     | This issue is related to our Project Fuse, which aims to speed up the site for end users                                                                                                                             |
+| Copy update 📝      | This issue is about a copy update, no design, UX or code changes                                                                                                                                                     |
```
pmahnke:
```suggestion
```

remove

cristinadresch:
@pmahnke could you progress this?

PR #204, opened 2019-10-18
Add ports for maas-ui

huwshimi:
Multiple ports are required by the maas-ui project.


jpmartinspt:
@huwshimi @squidsoup Do we need to open all these ports in the container for the UI to be usable?

huwshimi:
@jpmartinspt only one port is exposed, the other two are proxied to inside the container, but we can expose those individually as needed.

jpmartinspt:
@huwshimi So basically you might need to exposed them for dev purposes? If that's the case fix the linting and merge ahead!

huwshimi:
@jpmartinspt Yep that's right! Sometimes we'll need to run one of the servers in isolation.

jpmartinspt (approved):

huwshimi:
Thanks @jpmartinspt!

PR #205, opened 2019-10-27
Switch to YYL naming for iterations

matthewpaulthomas:
Currently, most squads name their iterations with the last two digits of the year, then the [week number](https://en.wikipedia.org/wiki/Week#Week_numbering) of the full week in the iteration. For example, the week starting 2019-10-28 is W44, so the iteration containing that week is “19-44”.

Since we have two-week iterations, iteration numbers increment by two. And as far as I know, week numbers aren’t common in _any_ of the countries where our team works, and certainly not in most of them. These two factors make the iteration numbers needlessly hard to remember.

Since the start of this year, in my own “squad” I’ve used a different scheme, YYL. 2019’s first iteration was 19A, the second 19B, and so on. Twenty-six letters in the Latin alphabet are enough for a full year of iterations. Advantages:
- Most importantly, it’s easier to spell and pronounce.
- No artificial pressure to have a one-week iteration at the start of the year. For example, if next January most people won’t start back until Monday 6th, we can just start iteration 20A on the 6th.
- No artificial pressure to have a one-week iteration before/after a cross-team sprint. For example, 19-20 was apparently a one-week iteration merely because the second week was a sprint. Using YYL, iteration 19J could just have paused for the duration of the sprint and resumed afterward.

Since it’s worked fine for the past ten months, I propose that we adopt this naming across the team.

### DIFF
--- a/project-management/iteration-creation.md
+++ b/project-management/iteration-creation.md
@@ -9,13 +9,13 @@ description: These are the steps to follow when creating a new iteration
 
 Go to your squad's repo and create a milestone titled:
 
-Iteration `{YY}`-`{week}` `{date from}` - `{date to}`
+Iteration `{YY}{letter}` ending `{YYYY-MM-DD}`
 
-For example: "Iteration 18-47 20th November - 3rd December."
+For example: “Iteration 19R ending 2019-11-07”
 
 ### Create a small task epic
 
-Create a new epic on the squad repo and attach it to the new milestone. Usually in the format "Small Tasks: Iteration {YY}-{week}"
+Create a new epic on the squad repo and attach it to the new milestone. Usually in the format “Small Tasks {YY}{letter}”
 
 ### Set up the backlog
 


anthonydillon (commented):
I do not like the numbered week notation myself as it is difficult to place historically and is clumsy to have to jump two at a time. One drawback with the proposed schema is that it does not intuitively allow for flexibility past 26 iterations. It's perceivable that we could switch to weekly iterations in certain circumstances. That does not mean this schema falls over, just that it requires more explanation of what a 30th iteration would be labelled as.

matthewpaulthomas:
If that ever happened, we could mimic spreadsheet columns: 20Y, 20Z, 20AA, 20AB, etc.

matthewpaulthomas:
> No artificial pressure to have a one-week iteration before/after a cross-team sprint.

This genuinely just happened:

@SirSamTumless: The next iteration will be one week long
Me: Why?
@SirSamTumless: Because next week we’re having a sprint
Me: That’s … not a reason. Why not just start the iteration the week after?
@SirSamTumless: Because … because then its name would be an odd number and I can’t handle it

nottrobin:
I like this idea. :+1:

nottrobin:
But we can't merge until I think at least @cristinadresch, @pmahnke, @cassiocassio are on board.

tbille:
I like this. :+1:  :+1:  :+1:  :+1:  :+1:  :+1:  :+1: 

Let's stop the numbers madness, I AM NOT A NUMBER!!!!

clagom:
To be honest I don't really think this would help. 52 weeks a year is a standard use and people can easily relate. 19-48 tells me exactly where in time this iteration is, 19-T does not.
That said, if everyone thinks it's better I guess we can adapt.

clagom:
> If that ever happened, we could mimic spreadsheet columns: 20Y, 20Z, 20AA, 20AB, etc.
E.g. this sounds really complicated and makes me wonder what a 20AA is.

matthewpaulthomas:
> E.g. this sounds really complicated and makes me wonder what a 20AA is.

I don’t think it’s reasonable to judge this proposal based on something (27+ iterations/year) that has never happened and isn’t likely to happen. If we had one-week iterations, week numbers would be fine.

hatched:
We have been naming our iterations as follows:
- Iteration 19-48 / Nov 25 - Dec 6
- Iteration 19-50 / Dec 9 - Dec 20
- Iteration 20-02 / Jan 6 - Jan 17
- Iteration 20-04 / Jan 20 - Jan 31

You'll note a few things:
- Every iteration is two weeks, even if one week of that is during a sprint, it doesn't matter as iteration work can be completed on the week of the sprint.
- The actual dates are included with the iteration name.
- People can refer to it by the date or the 'code' if they prefer.

nottrobin:
@hatched that's a good naming scheme - presumably it would fit in perfectly well with @matthewpaulthomas's suggestion in that the "code" would just change to the letter-based one.

I still like the letter-based scheme a lot better.

anthonydillon:
@SirSamTumless @cassiocassio @cristinadresch @pmahnke I think you need to take a stance on this proposal so we can decline or move forward with this one?

hatched:
FWIW we've been using the outline I've proposed above and it's been working quite well for us so far.

anthonydillon:
So this has now become problematic. The script that builds [the squads cycle report](https://docs.google.com/spreadsheets/d/14-tw6F7k964hvi2o-Ja3C3XhFNlnkq3_oeRNZbTkEA4/edit?usp=drivesdk) simply splits the master epics by the milestone the tasks were complete in. Therefore, we are getting iteration counts separated by the official method and this method when Matthew works on it. We have manually corrected that for now but it would be much better to not have too do that.

Matthew, would you be happy to concede this point for the benefit of the team as the idea generally had a split support?

matthewpaulthomas:
@anthonydillon My “squad” was merged into the Web squad at the end of last year. I have not created, and therefore not named, any iterations this year. So I don’t know what you mean by “when Matthew works on it”. It’s not up to me to decide, it’s up to Sam and Cristina.

nottrobin:
Closing for now.

PR #209, opened 2019-11-14
Team mantras

anthonydillon:
I have worked on outlining a number mantras I feel are important for the team to agree on when it comes to making UI decisions. Please take time to review the content and suggest improvements.

### DIFF
--- a/team/mantras.md
+++ b/team/mantras.md
@@ -1,0 +1,42 @@
+---
+title: Team mantras
+description: A set of high level mantras the web and design team embrace
+---
+
+## Never remove data or functionality because of screen size
+Never hide information or remove functionally on different screen sizes. In the modern era,  applications should be fully featured and easy to use on all devices and screens.
+
+**Do:** If a details page contains a set of metadata and actions which can be performed on that entity. That data and functionality should be accessible on all screens.
+For example, when viewing the list on a phone the user should be able to search and filter as they can on a desktop screen.
+
+**Don’t**: Reduce the information on the listing to fit the content in.
+For example, tables rows may need to stack or become expandible to persist that same information.
+
+
+## Do not assume a user's intent
+Any request for data should be clearly triggered by a user action such as a click or key press. Scrolling or hovering is not an intentional data request from a user.
+
+**Do**: Provide a button or interaction that will clearly request more data from the server.
+For example, provide a “Load more…” button at the end of a limited list to request to next page of data.
+
+**Don’t**: On mouseover or scroll request additional data from the server.
+For example: a tooltip that requests extra data to populate its contents.
+
+
+## Reflect state of the application in the URL
+All views in an application should be reflected in the URL so a user can share or bookmark an area and jump to it on load. This includes filter and search inputs on listing pages.
+
+**Do**:  When entering an entity details reflect that view in the URL.
+For example, /entity/12345
+
+**Don’t**: When clicking to reveal a popover update the address. For example, when filtering and selecting the filters from the drop down do not persist that state in the URL.
+For example, /listing/+filter-open
+
+
+## Always allow the user to undo an action
+If a user performs an action it should be clear how they can reverse the action.
+
+**Do**: Pop up a notification to undo the action or replace the action button with an opposite action.
+For example, Delete an entity and get a notification detailing its success and the action to undo.
+**Don’t**:
+For example, Mark an entity public without a clear action to make it private in case of mistakes.
--- a/team/mantras.md
+++ b/team/mantras.md
@@ -4,7 +4,8 @@ description: A set of high level mantras the web and design team embrace
 ---
 
 ## Never remove data or functionality because of screen size
-Never hide information or remove functionally on different screen sizes. In the modern era,  applications should be fully featured and easy to use on all devices and screens.
+
+Never hide information or remove functionally on different screen sizes. In the modern era, applications should be fully featured and easy to use on all devices and screens.
 
 **Do:** If a details page contains a set of metadata and actions which can be performed on that entity. That data and functionality should be accessible on all screens.
 For example, when viewing the list on a phone the user should be able to search and filter as they can on a desktop screen.
@@ -12,8 +13,8 @@ For example, when viewing the list on a phone the user should be able to search 
 **Don’t**: Reduce the information on the listing to fit the content in.
 For example, tables rows may need to stack or become expandible to persist that same information.
 
-
 ## Do not assume a user's intent
+
 Any request for data should be clearly triggered by a user action such as a click or key press. Scrolling or hovering is not an intentional data request from a user.
 
 **Do**: Provide a button or interaction that will clearly request more data from the server.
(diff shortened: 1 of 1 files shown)


team/mantras.md
```diff
@@ -0,0 +1,42 @@
...
+All views in an application should be reflected in the URL so a user can share or bookmark an area and jump to it on load. This includes filter and search inputs on listing pages.
+
+**Do**:  When entering an entity details reflect that view in the URL.
+For example, /entity/12345
+
+**Don’t**: When clicking to reveal a popover update the address. For example, when filtering and selecting the filters from the drop down do not persist that state in the URL.
+For example, /listing/+filter-open
+
+
+## Always allow the user to undo an action
```
squidsoup: Are these mantras retroactive? If so, this would require significant changes to both MAAS and RBAC (and possibly other apps of ours).

anthonydillon: Sure, as guidelines for all patterns and interactions going forward.

team/mantras.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Team mantras
+description: A set of high level mantras the web and design team embrace
+---
+
+## Never remove data or functionality because of screen size
```
nottrobin: For consistency, would it be better for each mantra to be communicated with the same force? As in, if others say "Do not", shouldn't we say "Do not" (or "Don't") here too? Never seems a bit too insistent.

team/mantras.md
```diff
@@ -0,0 +1,43 @@
...
+
+Any request for data should be clearly triggered by a user action such as a click or key press. Scrolling or hovering is not an intentional data request from a user.
+
+**Do**: Provide a button or interaction that will clearly request more data from the server.
+For example, provide a “Load more…” button at the end of a limited list to request to next page of data.
+
+**Don’t**: On mouseover or scroll request additional data from the server.
+For example: a tooltip that requests extra data to populate its contents.
+
+## Reflect state of the application in the URL
```
nottrobin:
```suggestion
## Reflect the state of the application in the URL
```

team/mantras.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Team mantras
+description: A set of high level mantras the web and design team embrace
+---
+
+## Never remove data or functionality because of screen size
+
+Never hide information or remove functionally on different screen sizes. In the modern era, applications should be fully featured and easy to use on all devices and screens.
```
nottrobin:
```suggestion
Avoid hiding information or functionally on different screen sizes. In the modern era, applications should be fully featured and easy to use on all devices and screens.
```

team/mantras.md
nottrobin: Yeah, I think it's fine for it to be a principle without us having successfully followed it everywhere. Going forward, we should make every effort to follow it, but there may still be (hopefully extremely rare) exceptions.

nottrobin:
Love these. :+1:. A few comments inline for your consideration, but I'm happy either way.

team/mantras.md
lilyvidenova: I don't think that this is always necessarily the best solution - I know this is going to escalate in a big debate, however I think you should aim to provide the correct amount of information/functionality for the context the user is in. More information/functionality does not always equal a better experience.

lilyvidenova:
I'm sorry if I sound dumb but I have no idea how to see the full file we are supposed to review.

squidsoup:
> I'm sorry if I sound dumb but I have no idea how to see the full file we are supposed to review.

You have to click the "Files changed" tab at the top:

image

lilyvidenova:
> > I'm sorry if I sound dumb but I have no idea how to see the full file we are supposed to review.
> 
> @lilyvidenova You have to click the "Files changed" tab at the top:
> 
> image

Thanks @squidsoup! It still took me a long time to figure out how to see the actual text from there. 

In more general terms - I think these are a good starting point but are too general. I also don't think this is a good way to have a discussion.

team/mantras.md
matthewpaulthomas: Undo is awesome (much better than a confirmation step), but sometimes it’s less awesome than other stuff we could implement in the same amount of time. So I think “Always” is too strong here.

Separately, many things can’t or shouldn’t be undoable. For example, formatting or repartitioning a disk, or deleting your PII from a service. One common substitute is providing a timeout. For example, you can’t undo sending e-mail, but in Gmail you get a few seconds to “Undo” (cancel) it instead. You can’t undo deleting your Facebook account, but you get 30 days to change your mind. And often you can’t undo deleting a file/message/photo, but instead items in a trash may expire after a week/month.

team/mantras.md
matthewpaulthomas:
```suggestion
Never make information or functions unavailable depending on screen size. Sites should be fully featured and easy to use on all devices and screens.
```

nottrobin:
@lilyvidenova what do you think is a better way?

There's Google Docs (which is how this started - https://docs.google.com/document/d/1ywsoF9Sp06hxLj4veC2VSHT2kPvJWHionOI1yeCgMAc/edit#) which give you better direct visibility over the document and make it easier to hold discussions on specific lines in comments. Using Google Docs is actually [suggested in CONTRIBUTING.md](https://github.com/canonical-web-and-design/practices/blob/master/CONTRIBUTING.md#collaborating-on-large-documents).

Is that what you're thinking of? Or do you think maybe email?

@clagom suggested (by email) a meeting during the planning week. While I think face-to-face discussions and meetings around larger practices are a great idea, and we should definitely do them, I don't think they should make up the entirety of the discussion. I think it's better to have a digital place to centralise the discussion (GitHub, A google doc, an email thread) which we can then have meetings around.

For me there are a number of problems with making all the decisions in the meeting:

- Some people may be absent
- Some people may not be listening
- Certain people tend to dominate discussion in meetings (e.g.: me), others don't say much
- There's often not time to consider other people's points properly and respond
- It's hard to bring evidence

So maybe we should move discussion back to the Google doc, and also schedule a meeting for the week after next, and anything that comes out of the meeting can also be used to update the Google doc. What do you all think?

team/mantras.md
nottrobin: I still think this is a good principle, and it can lead us to consider implementing interesting solutions like the Gmail delays for non-retractable / descructive actions like formatting.

Agree "always" is probably the wrong word.

team/mantras.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Do:** If a details page contains a set of metadata and actions which can be performed on that entity. That data and functionality should be accessible on all screens.
+For example, when viewing the list on a phone the user should be able to search and filter as they can on a desktop screen.
+
+**Don’t**: Reduce the information on the listing to fit the content in.
+For example, tables rows may need to stack or become expandible to persist that same information.
+
+## Do not assume a user's intent
+
+Any request for data should be clearly triggered by a user action such as a click or key press. Scrolling or hovering is not an intentional data request from a user.
```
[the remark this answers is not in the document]
nottrobin: I generally agree that we could maybe use slightly less strong language across the board.

Although I think it is interesting that this leads us to question image lazyloading. Another pattern would be to not show images until the user actually clicks to load them (like how flashblock works). In many cases I think that's probably the wrong solution, but it's interesting to be pushed to consider it.

team/mantras.md
nottrobin: Fair enough - I suppose I was saying I thought "never" was too strong wording in this case, but I'm open to being persuaded.

Especially, I think the "Never hide" in the next sentence is too strong - because as you suggest, one way to preserve functionality is to hide it behind a user action - click to expand, or a menu or something.

But I take your points. I think they're good.

@lilyvidenova if we expand on the point a bit to clarify that it's okay to *hide* functionality, as long as it's findable, do you still disagree? The point being that nothing you can do or can read on a desktop should be impossible to do / find on a phone?

I'd be interested to hear specific examples where you think it might be justified to remove information or functionality from phone users.

anthonydillon:
Sorry all, closing this PR and please move the conversation to [the google doc](https://docs.google.com/document/d/1ywsoF9Sp06hxLj4veC2VSHT2kPvJWHionOI1yeCgMAc/edit#).

PR #212, opened 2019-12-02
Add performance podcasts to suggested podcasts page

steverydz:
Two performance related podcasts have come out recently and I've found them interesting/informative so am proposing them here for our suggested podcasts list.


anthonydillon:
:+1:

barrymcgee:
@steverydz Run Prettier here to fix the build and it should be good to merge.

PR #213, opened 2019-12-03
Add advice for safely specifying files for npm packaging.

squidsoup:
## Context
Some of our projects are currently using `.npmignore`, which is not recommended by npm. `.npmignore` primarily exists to support projects that are *not* using git.


nottrobin:
:+1: looks good. Important point.

PR #216, opened 2020-01-09
Add tools page

solazio:
Please feel free to add any other tools you use.

### DIFF
--- a/coding/tools.md
+++ b/coding/tools.md
@@ -1,0 +1,19 @@
+---
+title: Tools we use
+description: List of tools we use to speed things up
+---
+
+This document outlines the tools we use to increase efficiency.
+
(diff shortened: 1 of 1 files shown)


deadlight:
Looks good to me. I'd say merge it then people can add other items as separate PRs. :+1:

steverydz:
I agree with @deadlight but don't tell him. Oh wait...

nottrobin:
If this is just a helpful list that we hope people will expand dynamically as and when they get a chance, I don't think the practices repo is quite the right place for it.

It's fine for now, but as soon as we have a more dynamic place for content (which I think we agreed would be discourse.ubuntu.com maybe?) we should move this list to there.

PR #219, opened 2020-01-23
Update editorconfig to use in all projects

barrymcgee:
@nottrobin Pls sanity check

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,7 +15,6 @@ charset = utf-8
 indent_style = space
 indent_size = 2
 
-# Matches multiple files with brace expansion notation
 [*.{py}]
 indent_style = space
 indent_size = 4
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,3 +1,16 @@
...
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace=true
+
+# Matches multiple files with brace expansion notation
+[*.{js,py,jinja2,html,json,md}]
```
squidsoup: Python uses 4 space indents, I think you should probably remove `py` here.

.editorconfig
```diff
@@ -1,3 +1,20 @@
...
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace=true
+
+# Matches multiple files with brace expansion notation
+[*.{js,jinja2,html,json,md}]
+charset = utf-8
```
nottrobin: No reason why this wouldn't apply to python too

PR #220, opened 2020-01-23
Reserve z-index ranges for plugins

nottrobin:
I was keeping this list of z-index ranges in [the wiki](https://github.com/canonical-web-and-design/practices/wiki).

However we're probably retiring the wiki in favour of Discourse, and
I think these ranges should really be in practices rather than Discourse,
if indeed it should exist at all.


hatched:
The z-index should be provided or at least overridable by the including project. I don't think it's really reasonable to "reserve" a z-index range.

nottrobin:
It's an interesting idea to make it an option on the module. We could explore that.

@anthonydillon - opinions?

anthonydillon:
Sounds like an interesting module for Vanilla but I would like to confirm if this is still a real issue that needs to be solved before tackling it.

nottrobin:
I think it's a thing that came up once, so I wrote it down in the wiki - but now I'm killing the wiki I feel it should either be a practice or not exist.

@anthonydillon Should I just close this PR for now? We can revisit it if we encounter clashing z-indexes again?

PR #228, opened 2020-03-17
Don't call it BEM

bartaz:
We call our stylesheet guidelines "BEM" (in the title), when strictly speaking we don't use BEM, only "BEM-inspired" naming convention.

This PR updates the title to not call the document "BEM standards", because our standards are not BEM.

### DIFF
--- a/coding/stylesheets.md
+++ b/coding/stylesheets.md
@@ -3,7 +3,6 @@ title: Stylesheet coding standards
 description: Code style standards for CSS and Sass
 ---
 
-
 Our stylesheet conding standard is based on [BEM methodology](http://getbem.com/).
 
 All class names in the codebase are built of three parts:
--- a/coding/stylesheets.md
+++ b/coding/stylesheets.md
@@ -1,5 +1,5 @@
 ---
-title: Stylesheets coding standard
+title: Stylesheet coding standards
 description: Code style standards for CSS and Sass
 ---
 
--- a/coding/stylesheets.md
+++ b/coding/stylesheets.md
@@ -3,22 +3,18 @@ title: Stylesheet coding standards
 description: Code style standards for CSS and Sass
 ---
 
-## BEM methodology
 
 Our stylesheet conding standard is based on [BEM methodology](http://getbem.com/).
 
-For a quick recap on BEM, it works by breaking down all classes in a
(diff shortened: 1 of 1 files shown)


coding/stylesheets.md
```diff
@@ -1,8 +1,12 @@
 ---
-title: BEM coding standard
+title: Stylesheets coding standard
```
nottrobin:
```suggestion
title: Stylesheet coding standards
```

coding/stylesheets.md
```diff
@@ -1,8 +1,12 @@
...
 description: Code style standards for CSS and Sass
 ---
 
+## BEM methodology
```
nottrobin: I think putting the BEM heading here is also too strong - again because these are our standard, not BEM's. Maybe leave this heading out so it just goes straight into the "introduction"? I would keep the line about "based on BEM..." but that seems quite sufficient.

coding/stylesheets.md
```diff
@@ -1,8 +1,12 @@
...
+
+Our stylesheet conding standard is based on [BEM methodology](http://getbem.com/).
+
 For a quick recap on BEM, it works by breaking down all classes in a
```
nottrobin: Again in the tone of taking ownership of our own guidelines, I think it might be nice to word this more like:

> All classes are broken down into one of three groups:

As I said, I think the "based on BEM methodology" in the line above should be quite sufficient.

bartaz: @nottrobin Thanks for comments. I updated the introduction text a bit.

nottrobin:
Cool thanks 👍 from me

coding/stylesheets.md
```diff
@@ -1,20 +1,19 @@
 ---
-title: BEM coding standard
+title: Stylesheet coding standards
 description: Code style standards for CSS and Sass
 ---
 
-For a quick recap on BEM, it works by breaking down all classes in a
-codebase into one of three groups:
+Our stylesheet conding standard is based on [BEM methodology](http://getbem.com/).
```
nottrobin: Oops. Should be

```suggestion
Our stylesheet coding standards are based on [BEM methodology](http://getbem.com/).
```

PR #236, opened 2020-06-04
Document JS testing attributes

anthonydillon:
Added a section about the naming syntax for testing attributes in JS.

### DIFF
--- a/coding/javascript.md
+++ b/coding/javascript.md
@@ -72,6 +72,10 @@ We use the [Jest](https://jestjs.io) unit test framework.
 - Reducers, action creators and selectors should all have simple functional tests.
 - Test composed Reselect selectors in isolation with [resultFunc](https://github.com/reduxjs/reselect#q-how-do-i-test-a-selector).
 
+##### Testing data-attr
+The naming convention for testing data attrubutes is: `data-test="<identifiable string>"`.
+This gives all projects a common interface when testing and search the project to testing attributes.
+
 ### Code style & formatting
 
 To ensure code formatting consistency across our codebase, we use [Prettier](https://github.com/prettier/prettier).


coding/javascript.md
```diff
@@ -72,6 +72,10 @@ We use the [Jest](https://jestjs.io) unit test framework.
 - Reducers, action creators and selectors should all have simple functional tests.
 - Test composed Reselect selectors in isolation with [resultFunc](https://github.com/reduxjs/reselect#q-how-do-i-test-a-selector).
 
+##### Testing data-attr
+The naming convention for testing data attrubutes is: `data-test="<identifiable string>"`.
+This gives all projects a common interface when testing and search the project to testing attributes.
```
hatched, line 77: This pattern isn't followed in the Juju Dashboard project where it's prefixed by `data-test` but then there are more detailed descriptions afterwards ex) `data-test-column="units"`

bartaz: I think what we discussed in catch-up is that as long as the attribute name starts with `data-test` it is fine, because it can be easily identified and searched for.

So both `data-test="something"` and `data-test-column="units"` should be OK.
We should just avoid any other custom stuff like `data-id=`, `data-column-test=`.

anthonydillon: I thought it would be easier to have a single rule which we can align on overtime. @hatched would you prefer to keep the flexibility to add wildcards to the data key?

hatched: I would, I was thinking how you could accomplish the same thing in the dashboard tests without it, and I can't think of anything else that would be more elegant so we should probably document that use case here.

anthonydillon:
Going to close this one as it seems we can't settle on a single naming convention.

PR #241, opened 2020-07-21
Add some initial advice for using TypeScript.

squidsoup:
Add a new *TypeScript* heading to the javascript doc, and provide some initial advice and examples of useful react patterns.

### DIFF
--- a/coding/javascript.md
+++ b/coding/javascript.md
@@ -8,24 +8,159 @@ redirect_from:
 
 This document outlines the rules for writing JavaScript across our codebase.
 
-## Progressive Enhancement
+## General Principles
+
+### Progressive Enhancement
 
 JavaScript should progressively enhance a page.
 
 On all publicly accessible websites, core content and functionality should still be available without JavaScript. JavaScript applications written for a more specific audience which won't be indexed by search engines may be more lax, but should still treat progressive enhancement as a guiding principle.
 
-## Performance
+### Performance
 
 - Cache DOM queries — only select an element once.
 - Use event delegation as much as possible to reduce the number of events bound to the page
 - Understand when you cause an element repaint and reduce inline style manipulations accordingly.
 - If possible, always use CSS for element transforms and transitions.
 
-## Element hooks
+### Element hooks
 
 We should target page element in JavaScript using the `data-js` element attribute instead of using the `class` attribute. Classnames should be used for styling only.
 
 e.g. `<a data-js="index-link" href="/index">Index</a>`.
+
+## TypeScript
+
+### TypeScript vs JavaScript
+
+For any medium-large JavaScript applications, consider using TypeScript in preference to JavaScript, as it offers a better developer experience, and increased safety. See [Transitioning to TypeScript](https://docs.google.com/document/d/1yXNI_biuKwoHBk2goLEksZayV_MGAkQxukvSNLCDtnE/edit#heading=h.yij8xaij9lcy) for more details.
+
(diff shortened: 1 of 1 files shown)


coding/javascript.md
```diff
@@ -8,25 +8,158 @@ redirect_from:
...
+
+`props` should be typed in the same file as the component, and typically a functional component will have the return type `JSX.Element`:
+
+```typescript
+type Props = {
+  name: string;
+  age: number;
+}
+
+const Person = ({name, age }: Props): JSX.Element => ()
```
Caleb-Ellis:
```suggestion
const Person = ({ name, age }: Props): JSX.Element => ()
```

coding/javascript.md
```diff
@@ -8,25 +8,158 @@ redirect_from:
...
+
+// Wheels.tsx
+export type Props = {
+  spinners: boolean;
+}
+
+export const Wheels = ({ spinners }: Props): JSX.Element => {
+  if (spinners) {
+    return (
+      <>My car is gangsta.</>
```
Caleb-Ellis, line 132: :motorized_wheelchair:

coding/javascript.md
```diff
@@ -8,25 +8,158 @@ redirect_from:
 
 This document outlines the rules for writing JavaScript across our codebase.
 
-## Progressive Enhancement
+## General Principals
```
bartaz, line 11: Is it "principals" or "principles"? Or are those the same?

coding/javascript.md
```diff
@@ -8,25 +8,158 @@ redirect_from:
...
+
+// Car.tsx
+import { Wheels } from "./Wheels";
+import type { Props as WheelsProps } from "./Wheels";
+
+type Props = {
+  colour: string;
+} & WheelsProps;
+
+const Car = ({colour, ...rest }: Props): JSX.Element => (
```
Caleb-Ellis, line 146:
```suggestion
const Car = ({ colour, ...rest }: Props): JSX.Element => (
```

coding/javascript.md
squidsoup: Oh, good catch, thanks :)

Caleb-Ellis: I read that three times and came to the conclusion that "principals" was indeed right, but it actually 100% should be "principles" hahaha

coding/javascript.md
```diff
@@ -8,25 +8,160 @@ redirect_from:
...
+For any medium-large JavaScript applications, consider using TypeScript in preference to JavaScript, as it offers a better developer experience, and increased safety. See [Transitioning to TypeScript](https://docs.google.com/document/d/1yXNI_biuKwoHBk2goLEksZayV_MGAkQxukvSNLCDtnE/edit#heading=h.yij8xaij9lcy) for more details.
+
+Remember that TypeScript gives you the ability to _progressively_ type your project; it isn't an all or nothing commitment. If you like, you can convert JavaScript files to TypeScript one at a time, if migrating an existing project.
+
+TypeScript is currently used in the following projects:
+
+- [maas-ui](https://github.com/canonical-web-and-design/react-components/)
+- [react-components](https://github.com/canonical-web-and-design/react-components/)
+
+### Patterns
```
bartaz, line 45: These are really helpful, thanks!

Maybe it would be also worth adding some links to real code examples?
Ideally for each pattern we could have "here is how we use it in this component" link. Or maybe just "Some real TS component examples" section with a couple examples?

squidsoup: Good suggestion! I can add a link to some real-world examples in `react-components`.

coding/javascript.md
hatched: An easy way to remember.... "a (school) principal is your PAL"  😄

coding/javascript.md
squidsoup: Done!

PR #256, opened 2021-07-12
Add Smashing podcast

mrgnr:
[Smashing](https://podcast.smashingmagazine.com/) covers lots of topics related to front-end development, React, frameworks and tools. I've found it helpful for getting a good overview of libraries and what problems they address.


anthonydillon:
Nice :+1:

tbille (approved):
nice, will give it a try. :headphones:

anthonydillon:
The guideline is to just :+1: if your happy with the amendment and to leave it open for at least 24 hours but in this case I think it's not contentious.

